PR #6, opened 2016-04-26
fix parsing of OS version number

halotron:
With default culture of mine, there was an exception for me in the ctor of RoboCommand, which this fixes by using a Convert.ToDouble using the invariant culture.


halotron:
will fix the fix... ;)

PR #7, opened 2016-04-26
fix OS version parsing issue due to default culture

halotron:
a fix for an issue when parsing a double using current default culture


tjscience:
Thanks halotron!

PR #9, opened 2016-04-28
Support of spaces in some properites

env123:
Source, Destination and FileFilter properties can contain spaces.

### DIFF
--- a/RoboSharp/CopyOptions.cs
+++ b/RoboSharp/CopyOptions.cs
@@ -1,4 +1,5 @@
-﻿using System.Text;
+﻿using System;
+using System.Text;
 
 namespace RoboSharp
 {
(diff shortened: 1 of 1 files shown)


tjscience:
Thanks for this! I cannot believe that I did not catch this before :o

mtudor:
I'm new to C# so I'm not quite at the stage where I can download, compile and create a pull request for this - hopefully I'll get there - but I suspect that this change might have broken the FileFilter when you want multiple wildcards (*.txt *.bmp).  On the RoboCopy command line, putting quotes around those wildcards causes it to be treated as a giant wildcard for a single file, rather than as separate wildcards in an OR configuration...

Can anyone confirm?

Issue #27, opened 2017-05-19
Support projects that target .NET 3.5


icnocop:
Hi.

Thank you for RoboSharp.

I'd like to use RoboSharp with my project, but my project targets .NET 3.5.

When I try to add the NuGet package, I get the error message `Could not install package 'RoboSharp 1.0.25'. You are trying to install this package into a project that targets '.NETFramework,Version=v3.5', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.`

It seems there are really only three incompatibilities in the current source code of RoboSharp that can be changed to make the code compatible with .NET 3.5:

1. `string.IsNullOrWhiteSpace`
Create an extension method instead:
```
public static class StringExtensions
{
    public static bool IsNullOrWhiteSpace(this string value)
    {
        if (value == null) return true;
        return string.IsNullOrEmpty(value.Trim());
    }
}
```

2. `Task`
Use the `TaskParallelLibrary` or `System.Threading.Tasks.Unofficial` NuGet packages instead.

See http://stackoverflow.com/a/22518228/90287

3. `Lazy<T>`
Here is a thread-safe implementation of Lazy<T> for .NET 3.5:
http://dannykendrick.blogspot.com/2012/10/lazy-net-35.html

Thank you.

PR #28, opened 2017-05-19
target .NET 3.5

icnocop:
Deleted automatically generated, non-source code, and user specific files
Fixes github issue #27


icnocop:
The build is currently failing because "nuget restore" isn't run as part of the (pre-) build process.

See https://www.appveyor.com/blog/2014/03/18/about-nuget-package-restore/

tjscience:
I have added nuget restore. Thanks

PR #34, opened 2017-10-18
Update RoboCommand.cs

### LINKED ISSUE #32: Calling Dispose in OnCommandCompleted() on RoboCommand instance results in Task Exception
[Edit] After running the code, the issue is, if you call dispose in the completed method, it will fire before the stop method is called at the end of the start task, leaving the process, not null, but no process associated with it. [/Edit]

image

bruceburge:
Added the Process.Exited Event, and a hasExited variable to address the issues of checking for Process.HasExited, when there is no longer an associated process, causing an InvalidOperationException.

This should resolve #32

### DIFF
--- a/RoboSharp/RoboCommand.cs
+++ b/RoboSharp/RoboCommand.cs
@@ -15,6 +15,7 @@ namespace RoboSharp
         private Process process;
         private Task backupTask;
         private bool hasError;
+        private bool hasExited;
         private bool isPaused;
         private CopyOptions copyOptions = new CopyOptions();
         private SelectionOptions selectionOptions = new SelectionOptions();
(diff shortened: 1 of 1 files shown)


RoboSharp/RoboCommand.cs
```diff
@@ -277,12 +279,18 @@ void process_ErrorDataReceived(object sender, DataReceivedEventArgs e)
                 OnCommandError(this, new ErrorEventArgs(e.Data));
             }
         }
-
+        
+	    void Process_Exited(object sender, System.EventArgs e)
+	    {
+		    hasExited = true;
+    	}
```
Mizipzor: You are mixing tabs and spaces in this block, seems the rest of the file sticks to spaces.

PR #35, opened 2017-12-08
Add a FileClassType enum to ProcessedFileInfo

ceffo:
This small change helps clients when processing `ProcessedFileInfo` differently based on the kind of information (file, new directory or system message).


bruceburge:
I think this would be really helpful as well, having to know the expected value of the strings is a little tough when you haven't used the code for a while.

PR #46, opened 2018-03-01
Set EnableRaisingEvents to trigger the "Exited" event.

ehrenfeu:
Otherwise the Process_Exited() handler introduced in commit baf933e
will not fire at all, thus leading to a wrong setting of "hasExited"
and (worse) making subsequent calls to Stop() stall indefinitely.

The "Exited" event not firing can easily be verified by adding a debug
message to the handler method (not done in this PR to prevent tainting
the code base).


bruceburge:
He's write, this was an oversight on my part, this should be accepted asap.

PR #59, opened 2018-04-19
Add interface so it can be used with ioc - spike for discussion.

generik0:
Please allow an interface so it is possible to register implemenation /service in ioc.
Also good for unit testing.

Now from dev, without .net standard


PCAssistSoftware:
Closed for now as all checks failed, no further discussion in the 3+ years since opened, and project has moved on / changed a lot since that time.  Not really sure of benefits, can be re-opened if OP comes back with more information / amended code.

PR #87, opened 2020-05-11
Added General Output event. Catches every details possible

whosenbocus:
Added General Output event - this trigger the event for every single verbose robocopy is generating. 
Nice to have when you want to collect every single information being generated.

Added Tab under BackupApp to show details of copy being generated from General Output event.

### DIFF
--- a/RoboSharp.BackupApp/MainWindow.xaml
+++ b/RoboSharp.BackupApp/MainWindow.xaml
@@ -225,6 +225,9 @@
             <TabItem Header="Errors" Name="ErrorsTab">
                 <DataGrid Name="ErrorGrid"></DataGrid>
             </TabItem>
+            <TabItem Header="Details" Name="DetailsTab">
+                <TextBlock Name="DetailsTextBlock"/>
+            </TabItem>
         </TabControl>
     </Grid>
 </Window>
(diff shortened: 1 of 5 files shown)


wingers999:
How is this any different than what already exists in RoboCopyResults.LogLines? this already contains all output and can be shown (as I do in my app) at end of backup?

whosenbocus:
This event will trigger whenever there is new output from robocopy. When transferring huge files over the network, which might take hours to complete, the new event will trigger whenever there is new verbiage - which can then be captured to screen/file.

With the RoboCopyResults.LogLines, you only have the information at the end of the backup - user would be in the unknown while file is being transferred. 

I would say RoboCopyResults.LogLines is output at the end while the event is progressive output throughout the transfer

wingers999:
"user would be in the unknown while file is being transferred."

You can show progress of a file copying using e.CurrentFileProgress so even if it is a large file copying across network you can show the progress of that file copying on a progress bar or just on a label

You can also use same method to show overall copy progress which is what I do in my GUI so I have two progress bars, one showing e.CurrentFileProgress and another using my own code which works out progress of total backup job and shows as a percentage

whosenbocus:
I guess this is based on usage.

It works in your scenario, but the change I brought is more for when you are transferring large file as part of a bigger process and outputting everything to flat file (where we cannot use the progress bar and neither update any UI) - we can only rely on the output being generated progressively and update the file.

wingers999:
Yes I run RoboSharp / RoboCopy initially with the /l switch (ListOnly) so I can work out from RoboCopyResults total number of files to copy/delete, and total size etc, then I can use this when running without the switch to calculate overall percentage and file counts to show in GUI

Fair enough, just explaining my usage and what we had already recently added

PR #90, opened 2020-10-06
Add support for /IS and /IT copy options

mrpeardotnet:
Add support for /IS and /IT copy options (include the same files, include the modified files)


wingers999:
Thanks for this, just got a couple of other major jobs on at present, but in next few days I will take a proper look and if okay merge your commits to the dev branch :)

wingers999 (commented):
Hi

Just found a few moments to check this and Include_Same /IS and Include_Tweaked /IT are already included in RoboSharp, they can be found in SelectionOptions.cs - see lines 20/21, 97-106 and 197-200

mrpeardotnet:
Ah sorry, missed that it is in SelectionOptions.cs instead of CopyOptions.cs. Thank you for clarifiaction and your effort ;).

PR #99, opened 2021-07-30
Wrap log file paths in quotes if needed

RFBomb:
provides a function to wrap the log paths in quotes (required if user hasn't already done so in their string and if the path has spaces in it).
Without this, if a path with spaces is supplied, the task will silently fail


PCAssistSoftware:
Thanks Robert - now merged for you

Issue #104, opened 2021-11-03
Stop() function doesn't work.


RFBomb:
per this thread:
https://stackoverflow.com/questions/56349033/robocopy-rh-run-end-time-parameter

The RunHours parameter is used to define a set time frame to run within. For example, only run between 2pm and 4pm.
I'm a bit confused as to why this restriction made it into the API.

as in, why use:

`if (process != null && CopyOptions.RunHours.IsNullOrWhiteSpace() && !hasExited)`

instead of 

`if (process != null && !hasExited)` 

Obviously I can do a workaround in the code by supplying it with a DateTime.Now() and format the string accordingly, but that seems obtuse. 

The secondary question here, is if RoboCommand.Dispose() is called, that should end the process, correct?
 this is how it is in the code:

```
if (process != null)
     process.Dispose();
```

Note that process.kill() is missing. 
I tried using RoboCommand.Dispose(), and it just kept running (even updating my progress bar). 

----------

Current Code for RoboCommand.Stop()

```
public void Stop()
{
   if (process != null && CopyOptions.RunHours.IsNullOrWhiteSpace() && !hasExited)   
   {
      process.Kill();
      hasExited = true;
      process.Dispose();
      process = null;
      isCancelled = true;
      isRunning = false;
   }
}
```

RFBomb:
Yesterday was a long day. I have my thinking reversed. 
I'm still wondering why the check for RunHours exists, but the code is supposed to stop it if RunHours is NOT SET. 
So my wrapper now sets the string to null, then uses  the Stop() command.

But, even so,  this Stop() command does not appear to be working. 

```
            RoboSharpObject.CopyOptions.RunHours = null;
            RoboSharpObject.Stop();
            if (RoboSharpObject.IsRunning)
                throw new Exception ("RoboSharp Command Still  Running!");
```

Throws the error every time.

Interestringly, the Pause() and Resume() methods work great.

Issue #73, opened 2019-01-16
Nuget Documentation


splendid25:
Hi,

There is no documentation in the nuget package, but I see the source code is well documented.
I think it's just a matter of embedding the XML in the nuget ?

RFBomb:
I too have noted this issue before. 

This thread has the fix for it: 

https://stackoverflow.com/questions/43305578/show-comments-in-nuget-package

https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/x4sa0ak0(v=vs.100)?redirectedfrom=MSDN

It seems to be just a simple checkbox isn’t being ticked in the compiler options page.

PCAssistSoftware:
Invalid path for XML documentation file in RoboSharp.csproj - it is looking for it in "C:\Users\rbrenckman\source\repos\RobotSoftwareTool_.Net Framework\RoboSharp\RoboSharp\RoboSharp.xml"

RFBomb:
Yea that’s my bad. I tried fixing that multiple times by making it
relative, but it wasn’t sticking. I had hoped that publishing it into the
nugget package would bundle it into the assembly.

The ‘Build Action’ property on the file likely has to be changed to
‘embedded resource’. Also, look at ‘Copy To Output Directory’

PCAssistSoftware:
Now looking... also think the below needs adding as per one of the above articles to csproj file as it seems using the checkbox makes it an absolute path and causes the issue

` <GenerateDocumentationFile>true</GenerateDocumentationFile>`

PCAssistSoftware:
Okay, it does seem that if you check the box "XML documentation file" on the build tab then for some reason it wrongly creates an absolute path as mentioned here - https://stackoverflow.com/questions/43305578/show-comments-in-nuget-package and here - https://developercommunity.visualstudio.com/t/default-xml-documentation-file-path-is-absolute/319886.

The solution is to NOT check the box but to manually edit the RoboSharp.csproj file and add the below:-

```
<PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net40|AnyCPU'">
    	<GenerateDocumentationFile>true</GenerateDocumentationFile>
  </PropertyGroup>
```

This seems to work as RoboSharp.XML is correctly copied to \RoboSharp-dev\RoboSharp-dev\RoboSharp\bin\Debug\net40 when building

[...]

Issue #102, opened 2021-10-31
IsRunning = True even if errors occur


JoE-Seeve:
If the source directory does not exist during a copy process, IsRunnung is set to True beforehand. Since the source directory is only checked after this, IsRunning must be set to False again in the event of an error. Since this is missing, IsRunning remains true even after the copy attempt has been completed.

PCAssistSoftware:
@JoE-Seeve - should be resolved by #107 - please confirm okay and will then close - thanks

JoE-Seeve:
I tested the changes with a non-existent source and target directory and found that IsRunning is not set to FALSE if the command is aborted due to an error.

In the original code (with testing source directory, and not testing the target directory), I just added one line here and it worked. But since the handling of IsRunning has changed, I can no longer really understand when IsRunning is set to FALSE.

            // make sure source path is valid
            if (!Directory.Exists(CopyOptions.Source))
            {
                Debugger.Instance.DebugMessage("The Source directory does not exist.");
                hasError = true;
                OnCommandError?.Invoke(this, new CommandErrorEventArgs("The Source directory does not exist."));
                Debugger.Instance.DebugMessage("RoboCommand execution stopped due to error.");
                tokenSource.Cancel(true);

                // Added by JoE - Running was set even if the process wasn't started due to file error
                isRunning = false;
                //
            }

PCAssistSoftware:
Thanks @JoE-Seeve - will noted - think we jsut need to add isRunning = false; after line 273 and 289 but will wait for others to comment in case changes to handling of isRunning I have missed

JoE-Seeve:
OK, but please have in mind that the code for testing source and target has changed and it seems like that additional lines with "IsRunning = False" are necessary.

PCAssistSoftware:
Yes I am fully aware, hence waiting for others to comment first

PCAssistSoftware:
@RFBomb - thoughts on this please?

RFBomb:
In #107, I had made a change to the STOP() function in order to more reliably set the  `isRunning` field. I had thought this would take care of this issue, but I now see it didn't (hence my note for needing testing).

What I hadn't noticed is the `cancellationToken.ThrowIfCancellationRequested();` call in the backup task factory section. Since this occurs before generating the backup task, the Stop() method never actually runs.

I think the cleanest way to ensure that isRunning is set when the cancellation token is requested here is putting this line above the `ThrowIfCancellation` line:

[...]

PR #107, opened 2021-11-04
Multiple Fixes

RFBomb:
Fixed Stop()  not working. 
#104  
Set flag `hasExited = false;` above the `process.Exited += Process_Exited;` command. 

The problem was I turned ListOnly ON, get the file count, set ListOnly OFF, then started copying. I then used the file count from ListOnly to incremement an overall progress bar. Running in succession like this means the `hasExited` flag was set on the first run, and not reset for the start of the second run. this fixes that. 

-----
#73 
Added Remarks for Public Consumption to most of the public items. Some items converted to Internal (RoboCopyResultsBuilder)
Also enabled the checkbox I noted in a comment on #73.

This generated RoboSharp/RoboSharp.xml.

----
Killing the Process during RoboCommand Disposal
Disposing of the process leaves it running by default. I'm unsure if other wanted to have this functionality left as is, so I created a new Property called `StopIfDisposing` (whose default value is false, so default functionality is unchanged). When set to true, the process will be killed during Disposal(bool). 
The idea here is that if the program exits, this can kill the robocopy commands generated by the program, instead of leaving them running in the background (thus killable only by task manager or command prompt)


PCAssistSoftware:
Excellent work :) - if @tjscience doesn't see it and merge it in next day or so then I will happily do it for you

RFBomb:
will have to be tested, but I think #102 is also fixed due to the change I made in the Stop() function:
on the first commit in this PR.  c5c400e11dc4166bf8f471bcc055b237e3e696c3

`isRunning = !hasExited;` is now outside of the IF block within the Stop method. (Instead of isRunning = false inside the if block)

RFBomb:
No clue why ce32daabd76e1627d3d0e85137460ee1d263abc8 had failed, but I broke the changes in that commit into 3 other commits and it passed.

tjscience (approved):
Nice work, thanks!

PCAssistSoftware:
Thanks for merging Terry :)

PCAssistSoftware:
@RFBomb - could you possibly give me an example of how you are using the changes made for statistics for average etc e.g. AddStatistic - just so I can get my head around its benefits and if I want to use it myself - thanks :)

PCAssistSoftware:
@RFBomb - could you possibly supply an example snippet to show how AddStatistic and AverageStatistic are used? - thanks :)

RFBomb:
I'll supply a few of my methods.

This method is within my wrapper class.

```
///<summary>Unset the event and store the <see cref="LoggingOptions.ListOnly"/> setting. Then count the files. Restore original settings afterward. </summary> 
        public async Task CountFilesAsync(string domain = "", string username = "", string password = "")
        {
            bool _list = this.LoggingOptions.ListOnly;
            bool _event = IsOverallProgressUpdaterSetup;
            bool _logging = this.LoggingOptions.CreateLogFile;
            this.LoggingOptions.CreateLogFile = false;
            this.LoggingOptions.ListOnly = true;
            this._CountFilesResults = await StartAsync(domain, username, password);
            this._FileCount = _CountFilesResults.FilesStatistic.Total;
            this.LoggingOptions.ListOnly = _list;
            this.LoggingOptions.CreateLogFile = _logging;
            this.IsOverallProgressUpdaterSetup = _event;
            return;
        }
``````

The remaining methods are housed in a custom control that I load whenever I start a new copy operation. 

This first method is creating a new task that essentially watches primary Run() thread, which is shown below this method. This allows the caller to await the robocopy operations. Basically doing the same thing as the ContinueWith task in RoboCopy.start, but implemented slightly differently. The control uses stop/start/pause/cancel buttons to interact with the main task and this task. (this is subject to change, as the project is still in heavy dev, and frankly my first time working heavily threaded)

```
/// <summary>
        /// Wait until the copy operations have finished running <br/>
        /// Waiting thread will check every 250ms until the tasks are complete or cancelled.
        /// </summary>
        /// <returns></returns>
        public Task WaitUntilComplete()
        {
            //Create new task to monitor this object
            Task t =  new Task( () =>
            {
                while (!IsComplete && !IsCanceled && !IsDisposed )
                    System.Threading.Thread.Sleep(250);
                return;
            });
            //Wait for the task to start before returning it to the caller
            t.Start();
            while (t.Status < TaskStatus.Running)
                System.Threading.Thread.Sleep(250);
            return t;
        }
```

The CustomControl loads up two progress bars and a listbox to display the logging results. What I'm actually doing is generating several robocopy tasks, then passing them in a list into the window itself. Once the window is displayed, it will call this Run() method, which kicks off the copy operation. 
- The ListBox has an item added to it each time the OnFileProcessed event fires.
- Since this is several robocopy commands running consecutively, I am using the  adding method to add the statistics to each other for generating an overall file count. 

The average method was mostly added in because I figured "well why not? thats something someone might do in a situation like this. Might as well have it baked in." I was mostly thinking about the throughput ( MB/S ) stat when I added that one, since adding that makes little sense.

```
private async Task Run()
        {
            //Run all RoboCopy Command at same time to get the total file counts for each one
            started = true;    //these are feeding properties of the window so the caller can interact with it
            cancelled = false;
            IsComplete = false;
            List<Task> tasks = new List<Task>();
            RoboSharp.Results.Statistic FileStats = new RoboSharp.Results.Statistic();
            RoboSharp.Results.Statistic DirStats= new RoboSharp.Results.Statistic();
            this.txt_Source.Text = "Scanning Files";
            this.txt_Destination.Text = "Please Wait";
            foreach (RoboCopy cmd in RoboCopyCommands)
            {
                ActiveCommand = cmd;
                cmd.RoboSharpObject.OnCommandError += RoboSharpObject_OnCommandError;
                cmd.RoboSharpObject.OnError += RoboSharpObject_OnError;
                cmd.CopyOptions.Threads = 3;
                tasks.Add(cmd.CountFilesAsync());
            }
            await Task.WhenAll(tasks); // Wait for all scanning operations  to finish asyncronously
            this.TotalFiles = 0;
            foreach (RoboCopy cmd in RoboCopyCommands)
                TotalFiles += cmd.TotalFiles;
            
            //Start copying files -> these run only one command since file sizes are unknown and one command at a time should provide best throughput. 
            //Also, most reliable logging in case extras / newer files exists.
            foreach (RoboCopy cmd in RoboCopyCommands)
            {
                ActiveCommand = cmd;
                this.txt_Source.Text = cmd.CopyOptions.Source;
                this.txt_Destination.Text = cmd.CopyOptions.Destination;
                //cmd.Create_ProgressBarUpdater(this.ProgressBar_1, true, HideProgressBarWhenComplete: false);
                cmd.RoboSharpObject.OnFileProcessed += RoboSharpObject_OnFileProcessed;
                cmd.RoboSharpObject.OnCopyProgressChanged += RoboSharpObject_OnCopyProgressChanged;
                cmd.RoboSharpObject.OnCommandCompleted += RoboSharpObject_OnCommandCompleted;
                cmd.CopyOptions.Threads = 1;
                RoboSharp.Results.RoboCopyResults results = await cmd.StartAsync();
                FileStats.AddStatistic(results.FilesStatistic);
                DirStats.AddStatistic(results.DirectoriesStatistic);
                if (!cmd.IsCancelled) Modules.Core.Logging.LogMain("Total Files Copied: " + results?.FilesStatistic?.Copied.ToString());
                if (cancelled) break;
            }
            started = false;
            IsComplete = true;
            SetPauseResumeButtonState(true);
        }
```

**Comment by PCAssistSoftware** [comment]
Fantastic, thanks @RFBomb - very handy and now I see your examples I fully understand the idea - I think this may well be quite useful for me too :)

**Comment by RFBomb** [comment]
In retrospect here, I realized that I added Average to the standard statistics, and not to the speed statistics ><

**Comment by RFBomb** [comment]
@PCAssistSoftware 
I'm gonna submit another PR to put the average into the Speed Statistics. Does it make sense to hide the average for the file stats, or leave as is?

**Comment by PCAssistSoftware** [comment]
@RFBomb - okay good idea - I would leave it as is for file stats, it is there if anyone wants to use it

**Comment by PCAssistSoftware** [comment]
@RFBomb  - thanks again for your methods above, have just spent some time going through them and it is very useful to see how you are doing things.  

Might be worth getting together some useful bits like this and also perhaps comparing how you, I and @JoE-Seeve are handling the showing / calculation of file progress and overall progress in our apps, so the readme can be updated with more useful information for users of the library.

**Comment by PCAssistSoftware** [comment]
P.S. Be interested to follow your progress as your app developes - as like the idea of running multiple "jobs" at once etc, I think @JoE-Seeve has done something similar perhaps using RoboCopy Jobs (/JOB)

**Comment by RFBomb** [comment]
My application is taking files from several Network locations and pulling
them down locally for a file batch. The files picked will vary heavily
depending on the settings the user has provided. Each folder has about 30
sub folders that need to be merged for the batch.

So I created an object to represent the network folder, with its
corresponding local cache. This will check to sync the network to local if
local is out of date. A new command will be generated to copy local to
destination. If the local needed updating that command is generated and
runs first.

All these commands are based off user selection in an options window. Once
the user says go, these are generated into a list, which is passed to the
user control.

**Comment by PCAssistSoftware** [comment]
@RFBomb - I notice in the example code you kindly provided above you have "cmd.CopyOptions.Threads = 3;" - is .Threads something you have added extra to the standard RoboSharp?  Just getting my head around idea of running multiple jobs as think is something I can really make use of

**Comment by PCAssistSoftware** [comment]
@RFBomb - question / suggestion - as it is we can use your AddStatistic method to add specific sets of statistics e.g.  FileStatistics and DirStatistics to a new statistics object

```
RoboSharp.Results.Statistic FileStats = new RoboSharp.Results.Statistic();
RoboSharp.Results.Statistic DirStats= new RoboSharp.Results.Statistic();

RoboSharp.Results.RoboCopyResults results = await cmd.StartAsync();
FileStats.AddStatistic(results.FilesStatistic);
DirStats.AddStatistic(results.DirectoriesStatistic);
```

What might be handy is if we could add ALL the results in one go rather than having to specifiy FilesStatistic, DirectoriesStatistic, SpeedStatistic or BytesStatistic

e.g.

```
RoboSharp.Results.Statistic MyStats = new RoboSharp.Results.Statistic();

RoboSharp.Results.RoboCopyResults results = await cmd.StartAsync();
MyStats.AddStatistic(results);
```

Thoughts?

**Comment by RFBomb** [comment]
Having that functionality would make sense, and should be fairly easy to implement.

**Comment by PCAssistSoftware** [comment]
@RFBomb - thanks, could you possibly provide me a small snippet on how to use .AverageStatistic for SpeedStatistic as can't get it working here so obviously missing something obvious - once I have it working then I will update readme.md with sample for both this and .AddStatistic

**Comment by RFBomb** [comment]
I hadn't actually used it, I just wrote it up per request. 
The way its written, you should be able to run two commands, and grab the results from each one. 
Then you either call the static method to return a new object with the averaged results, OR you pick one you want to store the average in and run the instance method, setting the secondary one as the one to average with. 

Since the object average method combines the results and divides by listCount+1, if you did the following, you would get an incorrect average. This is because the speed of the new object is zero, meaning you are averaging a speed of 0 with the actual speeds. In this case, dividing by 3 instead of by 2. 

```
// AVERAGE IS DIVIDED BY 3 INSTEAD OF 2
SpeedStatistic avg = new SpeedStatistic();
avg.AverageStatistic(Result1);
avg.AverageStatistic(Result2);
```
OR 
```
// AVERAGE IS DIVIDED BY 3 INSTEAD OF 2
avg.AverageStatistic( new SpeedStatistic[] { Result1, Result2} );
```

You would want to do either of the two methods below:

```
//Call Static Method to return new object with the average
SpeedStatistic avg = SpeedStatistic.AverageStatistics( new SpeedStatistic[] { Result1, Result2} );
```

OR 
```
//Result1 will now store the average statistic value. Result 2 can be disposed or or re-used for additional RoboCopy commands.
Result1.AverageStatistic(Result2);
```

**Comment by RFBomb** [comment]
I have submitted PR #113 to introduce the method to combine the results objects. 
I was a bit unsure how you wanted to do that, so it may need some tweaking. Specifically on the ResultsStatus object. 

Since that PR uses all of these new methods, lets continue this conversation on the PR.

**Comment by PCAssistSoftware** [comment]
Thanks, will reply on #113 - re my other question above - "I notice in the example code you kindly provided above you have "cmd.CopyOptions.Threads = 3;" - is .Threads something you have added extra to the standard RoboSharp? Just getting my head around idea of running multiple jobs as think is something I can really make use of"

**Comment by RFBomb** [comment]
Sorry, that was part of my wrapper class. This wrapper was prior to implementing the xml documentation, so I had basically duplicated the robosharp object structure in my wrapper, renaming some items for shorthand.

```

            ///<summary> Number of robocopy threads to spawn. Set to 0 to disable multi-threaded copying. </summary>
            public int Threads
            {
                get => Parent.RoboSharpObject.CopyOptions.MultiThreadedCopiesCount;
                set
                {
                    int val = (value < 0) ? 0 : (value > 10) ? 10 : value;
                    Parent.RoboSharpObject.CopyOptions.MultiThreadedCopiesCount = val;
                }
            }
```

**Comment by PCAssistSoftware** [comment]
I did wonder if it was just the usual thread related switch but just wanted to check in case missing anything useful :)

Just tested SpeedStatistic.AverageStatistics as per you examples and now have it working perfectly - thank you

**Comment by PCAssistSoftware** [comment]
@RFBomb 

Quick question if you don't mind - asking here as related to AverageStatistics for Speed

Just playing with running multiple jobs to work out my preferred method and the code below works really well for me (and in future can hopefully be used with JobOptions idea from other discussion to populate job values / switches etc)

AddStatistic for File, Directories and Bytes is perfect.

The only issue I am having is how to get AverageStatistics for SpeedStatistic when doing it this way - probably obvious but having brain freeze I think.

I know I can calculate it manually e.g. add up BytesPerSec from each Job and then divided it by JobList.Count - but would prefer to utilize the methods you have included if possible

Thoughts?

```
Dim FileStats As New RoboSharp.Results.Statistic()
Dim DirStats As New RoboSharp.Results.Statistic()
Dim ByteStats As New RoboSharp.Results.Statistic()

Dim AverageSpeedStats As Results.SpeedStatistic

Dim JobList As List(Of RoboCommand)

Private Async Sub Button5_Click(sender As Object, e As EventArgs) Handles Button5.Click

        JobList = New List(Of RoboCommand)

        Dim job1 As New RoboCommand
        job1.CopyOptions.Source = "C:\Users\darre\Desktop\1a"
        job1.CopyOptions.Destination = "C:\Users\darre\Desktop\2"
        job1.CopyOptions.CopySubdirectoriesIncludingEmpty = True

        Dim job2 As New RoboCommand
        job2.CopyOptions.Source = "C:\Users\darre\Desktop\1b"
        job2.CopyOptions.Destination = "C:\Users\darre\Desktop\2"
        job2.CopyOptions.CopySubdirectoriesIncludingEmpty = True

        JobList.Add(job1)
        JobList.Add(job2)

        For Each Job In JobList

            AddHandler Job.OnFileProcessed, AddressOf analyze_OnFileProcessed
            AddHandler Job.OnCommandCompleted, AddressOf analyze_OnCommandCompleted

            Dim jobresults As RoboSharp.Results.RoboCopyResults = Await Job.StartAsync()

            FileStats.AddStatistic(jobresults.FilesStatistic)
            DirStats.AddStatistic(jobresults.DirectoriesStatistic)
            ByteStats.AddStatistic(jobresults.BytesStatistic)

            If jobresults.SpeedStatistic IsNot Nothing Then
                    ' ????????????
            End If

        Next

End Sub
```

**Comment by PCAssistSoftware** [comment]
**Edit: original problem posted below removed as found cause was me being an idiot! :)**

Hi @RFBomb 

Further to my question above re speed statistics - which I still need to work out.

I have been continuing my testing of run multiple jobs using code as above, and also using my own custom class rather than a list to store the "jobs" so I can then store the loglines and status for each job with its corresponding robocommand information e.g.

 ```
Dim RoboSharpJobsList As New List(Of RoboSharpJobs)

        Dim job1 As New RoboCommand
        job1.CopyOptions.Source = "C:\Users\darre\Desktop\1a"
        job1.CopyOptions.Destination = "C:\Users\darre\Desktop\2"
        job1.CopyOptions.CopySubdirectoriesIncludingEmpty = True

        Dim job2 As New RoboCommand
        job2.CopyOptions.Source = "C:\Users\darre\Desktop\1b"
        job2.CopyOptions.Destination = "C:\Users\darre\Desktop\2"
        job2.CopyOptions.CopySubdirectoriesIncludingEmpty = True

        RoboSharpJobsList.Add(New RoboSharpJobs With {.Name = "Job1", .Command = job1})
        RoboSharpJobsList.Add(New RoboSharpJobs With {.Name = "Job2", .Command = job2})

For Each Job In RoboSharpJobsList

            AddHandler Job.Command.OnFileProcessed, AddressOf analyze_OnFileProcessed
            AddHandler Job.Command.OnCommandCompleted, AddressOf analyze_OnCommandCompleted

            Dim jobresults As RoboSharp.Results.RoboCopyResults = Await Job.Command.StartAsync()

            FileStats.AddStatistic(jobresults.FilesStatistic)
            DirStats.AddStatistic(jobresults.DirectoriesStatistic)
            ByteStats.AddStatistic(jobresults.BytesStatistic)


            ' add exitcode
            Job.ExitCode = jobresults.Status

            ' add loglines
            Job.Log = jobresults.LogLines

Next

Class RoboSharpJobs
    Property Name As String
    Property Command As RoboCommand
    Property ExitCode As Results.RoboCopyExitStatus
    Property Log As String()
End Class

```

Running multiple jobs like this seems to work well and is promising for my requirements.

RFBomb:
Since the speed statistic is an average, you have to be aware of averaging in a Zero, which is the default value. This will throw off the results. SO, the speed static object uses for averaging should not be populated/instantiated  until after the first job has run, otherwise you get incorrect average. (3+3+0 /3 = 2 when you’d expect 3 since only two jobs have run)

What I would do fore the library itself is create a new class that simply holds a List<RoboCopyResults> object. Then, as jobs are completed, simply add the results to the list. 

Then, get-only properties can be added to the class that, when requested, while use the static methods to return a new stats objext of the appropriate type. 

This way, you don’t have to worry about creating multiple stats object for each stat, just one object that holds all the results and reports the totals upon request. Adding will not be performed until that result of all is requested, which should also be more efficient as far as the library is concerned.

PCAssistSoftware:
That does indeed make sense.  

For all other stats (files, directories and bytes) it works fine as is, it is just speed I can't get working using my current methods of running multiple jobs, but certainly follow the logic of the changes you suggest.  

Whether I am capable of making those changes is another thing!

RFBomb:
Should be pretty simple class I can mock up this week if I have time.

For your testing for now, dim the speed stat object, but leave it as null.
In your job loop, check if the object is null. If it is, set it to the
first job’s result. After that it will not be null, so an ELSE statement
should call the average method w it h the current jobs’ results. This
should accomplish the average you expect.

PCAssistSoftware:
@RFBomb 

Thanks - much appreciated :)

justind-dev:
@RFBomb wanted to say thanks as well. I was in the same boat as @PCAssistSoftware  in understanding the best way to use the speed stats using Robosharp in C#. Going to look at your suggestions as well.

PCAssistSoftware:
@justind-dev - just for info, the above discussion was for using speed stats when running "multiple jobs", if just doing single jobs then instructions for usage covered in https://github.com/tjscience/RoboSharp#readme

PCAssistSoftware:
> Should be pretty simple class I can mock up this week if I have time. For your testing for now, dim the speed stat object, but leave it as null. In your job loop, check if the object is null. If it is, set it to the first job’s result. After that it will not be null, so an ELSE statement should call the average method w it h the current jobs’ results. This should accomplish the average you expect.

Okay, so declaring my variable to store the speed stats

`Dim AverageSpeedStats As Results.SpeedStatistic`

And then below as suggested but not giving me correct results - I am obviously missing the point or doing this wrong here? Can live without for now as my other method works and still only at a design / testing stage.

```
 If AverageSpeedStats Is Nothing Then

                AverageSpeedStats = jobresults.SpeedStatistic

            Else

                AverageSpeedStats = Results.SpeedStatistic.AverageStatistics(New Results.SpeedStatistic() {jobresults.SpeedStatistic, AverageSpeedStats})

            End If
```

RFBomb:
I’ve created a new object on a new branch. I’ll upload it tomorrow after
some basic testing on my end to try and produce the results

PR #108, opened 2021-11-05
Fix #43 - Needs Verification before Merge

### LINKED ISSUE #43: Copy from/to remote server doesn't work [WITH SOLUTION]
in **"RoboCommand.cs" Line 188** you try to create the destination directory. But you are not authenticated on the other server.

I created a solution from Jon Cole(https://stackoverflow.com/questions/8901476/creating-windows-folders-using-impersonation?answertab=active#tab-top).

I attached the new RoboCommand file:
[RoboCommand.txt](https://github.com/tjscience/RoboSharp/files/1720124/RoboCommand.txt)

RFBomb:
#43 never supplied a pull request. I figured while I was at it, I'd take a peek at their fix, and if it was easy enough implement it. 

Their fix basically involved created a new class to pass credentials into Source/Destination checks. I have incorporated this fix into my changes proposed in the other PR. But I made this into a separate branch w/ separate PR because I have no idea if this will break things for some people, so would like someone to test this prior to merging into the published package.

The changes are actually quite small,  its basically wrapping the check in a using() command. The last commit has the finalized changes.` using (var impersonation = new ImpersonatedUser(username, domain, password)) { }`
9ddb4cf531414959e5eaa6a543b860893421948e


RFBomb:
No idea why the check  is failing, as it builds just fine in VS2019

PCAssistSoftware:
@RFBomb  - happy to go through and review these changes, test and then merge -  but a lot of the changes showing on "Files changed" tab are already implemented in #107 but might it be possible to go through this pull request and tidy it up so just showing pending changes to make it easier to review?  Thanks for the great work

RFBomb:
Superceded by #110

Issue #109, opened 2021-11-09
Problems with the interpretation of byte values in Statistics


JoE-Seeve:
It seems that the interpretation of file lengths for the statistical values ​​does not work correctly if RoboCopy returns not bytes but megabytes, gigabytes, etc. in the form of, for example, "3.854 g" (= gigabytes).

In the function "ParseTokenString" in "Statistics.cs" there are two points that can lead to errors:
- It is checked for upper case letters, although lower case letters are delivered as a unit (e.g. 3.876 g).
- To convert from gigabytes (e.g. 3,876 g) to bytes, the value must be multiplied by the factor (e.g. 1024 POW 3 - for giga). The function divides the factor by the value.

An error will not occur too often, as it appears that RoboCopy mostly returns byte values ​​and less often other units such as kilo-, mega-, giga-bytes.

PCAssistSoftware:
Unless I am missing something this shouldn't be a problem if /BYTES is being included as a parameter for [RoboCommand](https://github.com/tjscience/RoboSharp/blob/52744180a40829c6b21f86ec4bb31c9f480814eb/RoboSharp/RoboCommand.cs#L367) as all values from RoboCopy are returned as bytes and are then handled correctly by statistics.cs

JoE-Seeve:
Yes, @PCAssistSoftware, that is correct, but if you remove /BYTE as a fixed parameter again, the "ParseTokenString" function does a useful job. And secondly, RoboSharp remains open to integrate further parameters. Here it would be, for example, the JOB parameters, where RoboCopy demands that /QUIT must be at the end and not /BYTE (if you want to use the JOB parameters).

PCAssistSoftware:
/BYTE has to be used for progress information to be calculated correctly as well, and the whole statistics section was created, tested and added with that in mind.  Would need to spend more time looking into what you are saying to find out if it is worth changing the whole system or not, hopefully at some point we will get some comments / input from the contributor who created the statistics code, but it works well how it was designed (I worked with him at the time to test it throughly).  When I have some time I will look into what you are saying in more detail.   Certainly the ordering of the parameters can be changed I suspect so that /QUIT appears at end if using /JOB.

JoE-Seeve:
... I'm using RoboSharp with an added "JobOptions.cs" and it works like it should (of course with the changes mentioned above).

PCAssistSoftware:
[...]

PR #113, opened 2021-11-19
Introduce the Combine() method for RoboCopyResults

RFBomb:
The RoboCopyResultsStatus will have to be verified to work properly.

### DIFF
--- a/RoboSharp/Results/RoboCopyExitCodes.cs
+++ b/RoboSharp/Results/RoboCopyExitCodes.cs
@@ -34,5 +34,7 @@ namespace RoboSharp.Results
         /// Either a usage error or an error due to insufficient access privileges on the source or destination directorie
         /// </summary>
         SeriousErrorOccoured = 0x10,
+        /// <summary>The Robocopy process exited prior to completion</summary>
+        Cancelled = -1,
     }
 }
--- a/RoboSharp/Results/RoboCopyResults.cs
+++ b/RoboSharp/Results/RoboCopyResults.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Text;
 
 namespace RoboSharp.Results
 {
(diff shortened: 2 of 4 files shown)


PCAssistSoftware:
Reply in relation to #107

> I hadn't actually used it, I just wrote it up per request. 

Oh sorry, I assumed your wrote it as part of pull request, didn't realise requested by someone else.

Thanks for examples though, exactly what I needed - will test later and get readme.md updated with notes.

----

Got .AddStatistics method working fine - that is going to be very useful for running multiple jobs.

----

Will verify this merge later and reply in more detail - when I suggested being able to merge all statistics in one go I had totally forgotten about RoboCopyResults.Status - don't think we need to merge this as will cause confusions as each .Status relevant to just the one job rather than all jobs I would think, so just the 4 x statistics to merge (DirectoriesStatistic, FilesStatistic, BytesStatistic and SpeedStatistic) and not Status

PCAssistSoftware:
Okay have got the pull request locally on my machine so I can do some testing, can you provide a snippet as to how this will currently work and I will get testing and get back to you

RFBomb:
Yea, at the moment I have mocked in the Adding methods into my project for when I eventually get around to writing the log to disk. 

I didn't meant to put the Average() methods into Statistics, I just did it because I didn't think about the fact that SpeedStatistics was a totally different object at the time. I'm also not using SpeedStatistics.Average(), as its not really relevant to my use case. But figured that it would make sense to have if other people are running consecutive copy operations like I decided to do.

It may make sense to ignore any objects that have Speeds/Stats that value 0 as far as averages are concerned, since those could potentially throw off the average. Maybe in a later revision this should be passed as a boolean to let the user decide how to handle that. 

For RoboCopyResults.Add, it probably also makes sense to ignore the SpeedStatistic of any cancelled item, since I believe that will have a null return for the other objects. As I said, this will have to be gone through for testing various things, such as cancelling operations. 

Speaking of, I added the 'WasCancelled' property to both the Enum and the Status object, since I found that cancelling returns a -1 result, which was not handled anywhere. I wasn't sure how to properly represent it in the enum, so I did  my best for a quick PR. 

As far as merging Status is concerned, I really wasn't sure about that one. One on hand, it makes sense to have it merged to have an overall representation of how all the operations went. On the other, as you say, it can make it confusing because it now represents multiple runs. My thought on fixing that would be possibly creating a new object that implements a private List of Results, and exposes methods to add to, retrieve from, evaluate overall results of the list. This could also prove useful by saving processing time until the moment the items are evaluated. For example:


```
Class ResultsSummary {

public List<RoboCopyResults> results { get ;}= new List<RoboCopyResults>();

/// <summary> Return the sum of all BytesStatistics objects </summary>
public Statistics TotalBytesStatistic => RoboCopyResults.CombineResults(results, "byteStats");
//A new method to selectively add/average the results would be required here to avoid adding all every time when only ByteStats are requested for example


/// <summary> Return true if any of the jobs cancelled prior to completion  </summary>
public bool AnyCancelled => list.any( r => r.Status.WasCancelled);

}
```

This would be a bit involved to write up, which I don't have the time for at the moment, but would probably be the best way to resolve any issues.

RFBomb:
The test for this PR as its currently written would be something like:

```
RoboCopyResults R1 = await COMMAND_1
RoboCopyResults R2 = await COMMAND_2

R1.CombineResults(R2);  //Combine results of R2 into R1 object. R1 object now is combined result of R1+R2.

//Use the static method to create a new object with the merged results
RoboCopyResults MergedResults = RoboCopyResults.CombineResults(new RoboCopyResults[]{R1,R2});
```

RFBomb:
anyway, I don't think I'd actually merge this in until we collectively decide if we want to create that other class to support this idea of a merged result object

PCAssistSoftware:
Okay the more I think about this, as I sit here testing the code.  

1) I don't think we need to worry about combining speed results as what is the point, as you say only useful thing to do with speed results is to work out the average and we can already do that using SpeedStatistic.AverageStatistics - so think we can lose combining speed related info.

2) Also combining Status is going to just get confusing as each result is specific to an individual job rather than the whole run - and really don't want to over complicate it, personally I think I would check status for each "job/run" separately and then deal with it in my own code anyway to get overall result.

3) As for loglines - not sure if worth doing or not?  probably not - EDIT: in fact definitely not from quick testing as just ends up a mess of information with multiple titles, job headers and summaries - so again I would handle the logging for each job in my own code in same way as I would with Status

So what I was looking for earlier was a quick way of "adding" all the other "statistic" results e.g. FilesStatistic, DirectoriesStatistic and BytesStatistic, so instead of having to add each of the three separately e.g:-

```
RoboSharp.Results.Statistic FileStats = new RoboSharp.Results.Statistic();
RoboSharp.Results.Statistic DirStats= new RoboSharp.Results.Statistic();
RoboSharp.Results.Statistic ByteStats= new RoboSharp.Results.Statistic();

RoboSharp.Results.RoboCopyResults results = await cmd.StartAsync();

FileStats.AddStatistic(results.FilesStatistic);
DirStats.AddStatistic(results.DirectoriesStatistic);
ByteStats.AddStatistic(results.BytesStatistic);
```

We could also have ability to add all three of them all in one go e.g.

```
RoboSharp.Results.Statistic AllStats = new RoboSharp.Results.Statistic();

RoboSharp.Results.RoboCopyResults results = await cmd.StartAsync();

AllStats.AddStatistic(results);
```

So perhaps all that is actually needed is an overload or similar in Statistic.cs so if you pass just results it adds FilesStatistic, DirectoriesStatistic and BytesStatistic compared to when passing results.BytesStatistic then it just adds selected stats (e.g. bytes in this case)

Hope this make sense, and sorry I over complicated it in first place, but when you start going through it in code rather than just in mind it just doesn't make sense to worry about combining speed and status

RFBomb:
I agree with your points. I had simply thrown this together as proof of concept more than anything, knowing it should be messed around with and finalized. That is also where I think introducing an object to represent multiple consecutive runs may come in handy. As you could run a method like:

`ResultsSummary.AddResult(await COMMAND.StartAsync);`

Basically allow the user to initialize the result summary object, than as each consecutive command runs, the results can be easily added to the list. This allows best of both worlds: the ability to evaluate a bunch of jobs that ran consecutively as a whole, while also being able to access each result object in the list individually.

PCAssistSoftware:
Yes I like that idea, as you say gives the best of both worlds, being able to have "grouped" results for whole job, but also being able to dig down and look at results for individual runs.  Means I can run lots of little "copy" jobs and then analyze results at the end

PCAssistSoftware:
might be worth converting this pull request to draft whilst discussions and testing continue?

MarkECrum:
Guys, as I was implementing Job Options I notice there is no JobOptions model.... In reading, this thread do I understand this correctly... At this time our UI should handle job file logging???

RFBomb:
What do you mean, Job Options? What would the end goal be here? Set up an object that can hold multiple ROboCopy Commands or add new ones?
My project just creates a List<RoboCommand> and loops through it. This  Draft PR was more for a concept of combined results than anything, but a 'job' object to run multiple commands might be doable

MarkECrum:
Hi Robert,
On the microsoft site for the docs it shows Copy Options, File Selection
Options, Retry Options, Logging Options and then "Job Options" (list below
for clarity).
Now I believe that this is easily accomplished like you said, build a list
and loop it to the text file for results after processing... I'm not even
sure what the other options really do for you other than /job and /save.
In my UI I run through my check and text boxes to set the parameters.
Writing a bit of code to save the state of the boxes to a "Job" file and
then being able to load it is no problem. So I guess I was inquiring just
to verify that you had not incorporated those options in RoboSharp. Just a
note: I have not even run a job through it yet. I am now done with ALL of
the options in a nice tab interface and now I'm on the Main Tab with the
task at hand of running and reporting on the outcome. So I'm still unclear
how all of that works until I look at it. Today is the day for that.

I also want to thank you. I was going to write the exact same thing and ran
across RoboSharp. Very Cool!

Job options
Job options
Option Description
/job:<jobname> Specifies that parameters are to be derived from the named
job file.
/save:<jobname> Specifies that parameters are to be saved to the named job
file.
/quit Quits after processing command line (to view parameters).
/nosd Indicates that no source directory is specified.
/nodd Indicates that no destination directory is specified.
/if Includes the specified files.

PCAssistSoftware:
Hi @MarkECrum - another contributor @JoE-Seeve mentioned in issue linked below that he has implemented JobOptions.cs so may be worth getting input from him as well to acheive what you want - as I think it is something we could all make use of :), so very interested to follow progress and help test anything that comes up as a result of this :)

https://github.com/tjscience/RoboSharp/issues/109#issuecomment-970652699

JoE-Seeve:
Yes, I have implemented the job functions and they work as described in RoboCopy. In #109 I described how this also works for RoboSharp, but the changes that make sense for this (which do not affect existing applications) were rather rejected here.

RFBomb:
@JoE-Seeve, if I recall, the only suggestion that was being rejected from you was the removal of the /BYTE flag, since that flag is used to normalize the results reported by RoboCopy for things like event parsing. If you want JobOptions.cs included, submit a PR, since that on its own looked fine to me for merging. 

 [ Deleted comment -> Moved entirety of relevant deleted portion to #125 since its not relevant to this PR ]

PCAssistSoftware:
@RFBomb - yes that was my understanding of RoboCopy "Job" as well, mainly for loading and saving of jobs.  Nothing to do with running multiple jobs in sequence which is what I think a lot of us would like to be able to achieve.  Just not sure however if this is something that should be added to RoboSharp or is just something that should be done in our own GUI's / calling apps, but might be nice to have a proper method for being able to do so somehow.  Hope this reply make sense, typing on my phone...

RFBomb:
[ Deleted comment -> Moved entirety to #125 since its not really relevant to this PR ]

PCAssistSoftware:
@RFBomb - thanks, all very useful, will wait to see what others come up with.... :)

RFBomb:
I have truncated my comments in this thread since they were quite long due to the code i provided, and not relevant to this PR at all. 
@MarkECrum, @JoE-Seeve  Please move all discussion of JobOptions over to #125

RFBomb:
Superceded by #127

PR #115, opened 2021-12-01
Update StringExtensions class

RFBomb:
Convert StringExtensions to Internal class instead of public - prevent outside packages seeing this extension method if declared using Robosharp namespace
If public is still preferred, added appropriate xml summary tags.

I was having an issue in my project due to using the robosharp namespace, which enabled this extension method. I thought the problem was with the method itself, but took some time to find out where it was being pulled from. 

Two-fold solution: 
- Modify the protection level to avoid outside projects seeing the extension method
- Add in summary tags to describe the method and declare that its being provided via the RoboSharp namespace

### DIFF
--- a/RoboSharp/StringExtensions.cs
+++ b/RoboSharp/StringExtensions.cs
@@ -1,8 +1,13 @@
 ﻿namespace RoboSharp
 {
-    public static class StringExtensions
+    /// <summary>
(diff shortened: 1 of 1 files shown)


PCAssistSoftware:
Looks good to me, and a quick test here shows no problems - so merged :)

Issue #116, opened 2021-12-04
strange behavior if chosen source is ROOT of the same drive you are running RoboSharp from


PCAssistSoftware:
Never noticed this before and have tried it with older RoboSharp.dll just to ensure it wasn't recent changes and it wasn't :)

If I run my app (or even the example BackupApp) from for example my E: drive (located in e:\apps\robosharp - location not really relevant) and then choose E: as my source path it only lists / copies the files / folders in the root and doesn't do sub-folders etc.

If I run it again from C:\test and choose E:\ as my source then it works perfectly fine and does all directory levels.

Thought I was going mad, so tried it on another computer and also another drive - and in all cases if you are running RoboSharp from the same drive you are choosing as your source AND you are choosing the root of that drive as your source.

Example video may illustrate it better

https://www.screencast.com/t/9DR1BqNfHgS

First attempt is app running from E: and selecting E: as source - as you can see it only finds a few files / directories and if I check log it is just seeing root and NOT doing sub-directories

Second attempt is exactly same app running from C: and selecting E: as source - now works fine and does all levels

@RFBomb or @JoE-Seeve - can you replicate this?  Any thoughts?

JoE-Seeve:
OK, according to @PCAssistSoftware description I tried to make a copy with my app, with the BackupApp and with RoboCopy. My results were identical if I use RoboSharp. According to my tests, it is not a RoboCopy problem.

In my tests, only the program directory of the started program (my app or BackupApp) was copied.

@PCAssistSoftware: Please check the log file and here exactly the source directory. I could see that in this case the source directory points to the start directory of the copy program!

PCAssistSoftware:
> @PCAssistSoftware: Please check the log file and here exactly the source directory. I could see that in this case the source directory points to the start directory of the copy program!

Yes same here,  good spot - and thanks for testing for me - thought I was going mad!

@RFBomb @tjscience - any thoughts on this

JoE-Seeve:
I checked all the functions that change the source directory, because I noticed during a debug that "C: \\" results in "C:". In "ExtensionMethods.cs", the "CleanDirectoryPath" method deletes all backslashes that are at the end of a path. That is absolutely correct, but not if you already have the root directory here. I also don't think that a normal path with a backslash at the end would be wrong.

            // Get rid of trailing Directory Seperator Chars
[...]

PR #119, opened 2021-12-07
Add property to check for condition described by #116

RFBomb:
@PCAssistSoftware 

I added a property with a pretty detailed description. Can you test that my boolean logic is correct based on your results?
Also, that the explanation tags make sense (its a complex issue, so it needed a complex summary)

### DIFF
--- a/RoboSharp.BackupApp/MainWindow.xaml.cs
+++ b/RoboSharp.BackupApp/MainWindow.xaml.cs
@@ -114,6 +114,9 @@ namespace RoboSharp.BackupApp
             copy.LoggingOptions.NoFileSizes = NoFileSizes.IsChecked ?? false;
             copy.LoggingOptions.NoProgress = NoProgress.IsChecked ?? false;
 
+            //Removable Drive Fix
+            copy.SetWorkingDirectoryToSource();
+
             copy.Start();
         }
 
--- a/RoboSharp.BackupApp/MainWindow.xaml.cs
+++ b/RoboSharp.BackupApp/MainWindow.xaml.cs
@@ -114,9 +114,6 @@ namespace RoboSharp.BackupApp
             copy.LoggingOptions.NoFileSizes = NoFileSizes.IsChecked ?? false;
             copy.LoggingOptions.NoProgress = NoProgress.IsChecked ?? false;
 
-            // Working Directory / Root Source Conflict Fix
-            copy.SetWorkingDirectoryToSource();
-
             copy.Start();
         }
 
--- a/RoboSharp.BackupApp/MainWindow.xaml.cs
+++ b/RoboSharp.BackupApp/MainWindow.xaml.cs
@@ -114,7 +114,7 @@ namespace RoboSharp.BackupApp
             copy.LoggingOptions.NoFileSizes = NoFileSizes.IsChecked ?? false;
             copy.LoggingOptions.NoProgress = NoProgress.IsChecked ?? false;
 
-            //Removable Drive Fix
+            //Working Directory / Root Source Conflict Fix
             copy.SetWorkingDirectoryToSource();
 
             copy.Start();
--- a/RoboSharp.BackupApp/MainWindow.xaml.cs
+++ b/RoboSharp.BackupApp/MainWindow.xaml.cs
@@ -114,7 +114,7 @@ namespace RoboSharp.BackupApp
             copy.LoggingOptions.NoFileSizes = NoFileSizes.IsChecked ?? false;
             copy.LoggingOptions.NoProgress = NoProgress.IsChecked ?? false;
 
-            //Working Directory / Root Source Conflict Fix
+            // Working Directory / Root Source Conflict Fix
             copy.SetWorkingDirectoryToSource();
 
             copy.Start();
--- a/RoboSharp/CopyOptions.cs
+++ b/RoboSharp/CopyOptions.cs
@@ -282,6 +282,25 @@ namespace RoboSharp
 
         #endregion Public Properties
 
+        /// <summary>
+        /// Used by the Parse method to sanitize path for the command options.<br/>
+        /// Evaluate the path. If needed, wrap it in quotes.  <br/>
+        /// If the path ends in a DirectorySeperatorChar, santize it to work as expected. <br/>
+        /// </summary>
+        /// <param name="path"></param>
+        /// <returns>Each return string includes a space at the end of the string to seperate it from the next option variable.</returns>
+        private string WrapPath(string path)
+        {
+            if (!path.Contains(" ")) return $"{path} "; //No spaces, just return the path
+            //Below this line, the path contains a space, so it must be wrapped in quotes.
+            if (path.EndsWithDirectorySeperator()) return $"\"{path}.\" "; // Ends with a directory seperator - Requires a '.' to denote using that directory. ex: "F:\."
+            return $"\"{path}\" ";
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <returns></returns>
         internal string Parse()
         {
             Debugger.Instance.DebugMessage("Parsing CopyOptions...");
@@ -289,8 +308,8 @@ namespace RoboSharp
             var options = new StringBuilder();
 
             // Set Source and Destination
-            options.Append($"\"{Source}\" ");
-            options.Append($"\"{Destination}\" ");
+            options.Append(WrapPath(Source));
+            options.Append(WrapPath(Destination));
 
             // Set FileFilter
             // Quote each FileFilter item. The quotes are trimmed first to ensure that they are applied only once.
--- a/RoboSharp/CopyOptions.cs
+++ b/RoboSharp/CopyOptions.cs
@@ -298,7 +298,7 @@ namespace RoboSharp
         }
 
         /// <summary>
-        /// 
+        /// Parse the class properties and generate the command arguments
         /// </summary>
         /// <returns></returns>
         internal string Parse()
--- a/RoboSharp/ExtensionMethods.cs
+++ b/RoboSharp/ExtensionMethods.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using System.Text.RegularExpressions;
 
@@ -44,7 +45,7 @@ namespace RoboSharp
             }
 
             // Fix UNC paths that are the root directory of a UNC drive
-            if (new System.Uri(path).IsUnc)
+            if (Uri.TryCreate(path, UriKind.Absolute, out Uri URI) && URI.IsUnc)
             {
                 if (path.EndsWith("$"))
                 {
--- a/RoboSharp/ExtensionMethods.cs
+++ b/RoboSharp/ExtensionMethods.cs
@@ -1,4 +1,5 @@
 using System.IO;
+using System.Text.RegularExpressions;
 
 namespace RoboSharp
 {
@@ -17,8 +18,11 @@ namespace RoboSharp
         public static string CleanDirectoryPath(this string path)
         {
             // Get rid of single and double quotes
-            path = path.Replace("\"", "");
-            path = path.Replace("\'", "");
+            path = path?.Replace("\"", "");
+            path = path?.Replace("\'", "");
+
+            //Validate against null / empty strings. 
+            if (string.IsNullOrWhiteSpace(path)) return string.Empty;
 
             // Get rid of padding
             path = path.Trim();
@@ -28,6 +32,15 @@ namespace RoboSharp
             while (path.Length > 3 && path.EndsWithDirectorySeperator())
             {
                 path = path.Substring(0, path.Length - 1);
+            }
+
+            //Sanitize invalid paths -- Convert E: to E:\
+            if (path.Length <= 2)
+            {
+                if (DriveRootRegex.IsMatch(path))
+                    return path += '\\';
+                else
+                    return path;
             }
 
             // Fix UNC paths that are the root directory of a UNC drive
@@ -41,6 +54,8 @@ namespace RoboSharp
 
             return path;
         }
+
+        private static readonly Regex DriveRootRegex = new Regex("[A-Z]:", RegexOptions.Compiled);
 
         /// <summary>
         /// Check if the string ends with a directory seperator character
--- a/RoboSharp/ExtensionMethods.cs
+++ b/RoboSharp/ExtensionMethods.cs
@@ -2,7 +2,7 @@ using System.IO;
 
 namespace RoboSharp
 {
-    public static class ExtensionMethods
+    internal static class ExtensionMethods
     {
         public static string CleanOptionInput(this string option)
         {
@@ -23,8 +23,18 @@ namespace RoboSharp
             // Get rid of padding
             path = path.Trim();
 
+            return path.RemoveTrailingSlashes();
+        }
+
+        /// <summary>
+        ///  Get rid of trailing Directory Seperator Chars
+        /// </summary>
+        public static string RemoveTrailingSlashes(this string path)
+        {
+            if (string.IsNullOrWhiteSpace(path)) return string.Empty;
+
             // Get rid of trailing Directory Seperator Chars
-            while(path.Length > 0 && (path.EndsWith(Path.DirectorySeparatorChar.ToString()) || path.EndsWith(Path.AltDirectorySeparatorChar.ToString())))
+            while (path.Length > 0 && (path.EndsWith(Path.DirectorySeparatorChar.ToString()) || path.EndsWith(Path.AltDirectorySeparatorChar.ToString())))
             {
                 path = path.Substring(0, path.Length - 1);
             }
--- a/RoboSharp/ExtensionMethods.cs
+++ b/RoboSharp/ExtensionMethods.cs
@@ -39,7 +39,7 @@ namespace RoboSharp
             if (path.Length <= 2)
             {
                 if (DriveRootRegex.IsMatch(path))
-                    return path += '\\';
+                    return path.ToUpper() + '\\';
                 else
                     return path;
             }
@@ -56,7 +56,7 @@ namespace RoboSharp
             return path;
         }
 
-        private static readonly Regex DriveRootRegex = new Regex("[A-Z]:", RegexOptions.Compiled);
+        private static readonly Regex DriveRootRegex = new Regex("[A-Za-z]:", RegexOptions.Compiled);
 
         /// <summary>
         /// Check if the string ends with a directory seperator character
--- a/RoboSharp/ExtensionMethods.cs
+++ b/RoboSharp/ExtensionMethods.cs
@@ -23,23 +23,29 @@ namespace RoboSharp
             // Get rid of padding
             path = path.Trim();
 
-            return path.RemoveTrailingSlashes();
-        }
-
-        /// <summary>
-        ///  Get rid of trailing Directory Seperator Chars
-        /// </summary>
-        public static string RemoveTrailingSlashes(this string path)
-        {
-            if (string.IsNullOrWhiteSpace(path)) return string.Empty;
-
             // Get rid of trailing Directory Seperator Chars
-            while (path.Length > 0 && (path.EndsWith(Path.DirectorySeparatorChar.ToString()) || path.EndsWith(Path.AltDirectorySeparatorChar.ToString())))
+            // Length greater than 3 because E:\ is the shortest valid path
+            while (path.Length > 3 && path.EndsWithDirectorySeperator())
             {
                 path = path.Substring(0, path.Length - 1);
             }
 
+            // Fix UNC paths that are the root directory of a UNC drive
+            if (new System.Uri(path).IsUnc)
+            {
+                if (path.EndsWith("$"))
+                {
+                    path += '\\';
+                }
+            }
+
             return path;
         }
+
+        /// <summary>
+        /// Check if the string ends with a directory seperator character
+        /// </summary>
+        public static bool EndsWithDirectorySeperator(this string path) => path.EndsWith(Path.DirectorySeparatorChar.ToString()) || path.EndsWith(Path.AltDirectorySeparatorChar.ToString());
+
     }
 }
--- a/RoboSharp/RoboCommand.cs
+++ b/RoboSharp/RoboCommand.cs
@@ -447,6 +447,47 @@ namespace RoboSharp
                 parsedRetryOptions, parsedLoggingOptions);
         }
 
+        #region < Workaround for Issue #116  - WorkingDirectoryConflict Property >
+
+        /// <summary>
+        /// Check <see cref="CopyOptions.Source"/> to see if it is set to the root of the same drive of the current working directory <br/>
+        /// - If this occurs, robocopy will copy the working directory instead of the source directory. This is an issue with RoboCopy.exe itself. <br/>
+        /// - This will typically only occur on applications that are portable. Meaning they are installed onto a Removable Drive with the goal of copying files off the root of the same removable drive. <br/>
+        /// - As such, this property should able to be safely ignored if the app will not be running on a removable drive.
+        /// <para>
+        /// For example: <br/>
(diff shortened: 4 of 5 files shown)


PCAssistSoftware:
EDIT: comments removed as issue was with my build environment

RFBomb:
this PR did nothing (currently) to resolve the trailing backslash situation. It was only for the check to see if the working directory would be a conflict.

PCAssistSoftware:
EDIT: comments removed as issue was with my build environment

RFBomb:
what line?

PCAssistSoftware:
> Reference in n

Ignore me - my source I was making changes in seems to be corrupt as having other issues now - give me a few moments to download fresh and make change and i'll get back to you.  Sorry :)

PCAssistSoftware:
Okay ignore previous comments - issue was with my build environment - must of made a change somewhere I hadn't reverted.

Just run again with fresh source files.

My app is running from "E:\Programming\Visual Basic\Apps Developed\RoboCopy GUI\RoboCopy GUI"

Selected source as "E:"

Value of WorkingDirectoryConflictExists is False when should be True

PCAssistSoftware:
Okay, just done some debugging and I think it is the C# vs VB escape character / trailing slash issue at play here

In Backup App or using any C# app the value of CopyOptions.Source is `E:\\` which returns as `E:\` from Path.GetPathRoot

In my app using VB the value of CopyOptions.Source is `E:` which returns as `E: `from Path.GetRootPath

So the line 

`return Path.GetPathRoot(workingDir) == Path.GetPathRoot(CopyOptions.Source);`

is returning false as comparing E:\ to E:

RFBomb:
yea I just found that out myself while testing it.

couple strange things going on here. 
I am running the backup app under E:\debug\backupapp.exe
with a source of "E:"

the working directory == "E:\debug\"
`Path.GetFullPath(copyOptions.Source)  returns "E:\debug\" - was not expecting that!

so I'm debugging this check and working through it

PCAssistSoftware:
That is odd - like it has changed the Source path to the working directory prior to reaching that code?

Reference the \ difference between C# and VB - that certainly doesn't help for this or the other issue, as you can see in first image below the Source text box in Backup App contains E:\, but the value of Source.Text is E:\\

image

If I set text box to E: then it is returned as E:

In VB if text box contains E:\ then it is returned as E:\

RFBomb:
I believe the root issue you are chasing here is something within windows itself. 

Partially described here:
https://docs.microsoft.com/en-us/windows/win32/fileio/changing-the-current-directory

"
Although each process can have only one current directory, if the application switches volumes by using the SetCurrentDirectory function, the system remembers the last current path for each volume (drive letter). **This behavior will manifest itself only when specifying a drive letter without a fully qualified path when changing the current directory point of reference to a different volume. This applies to either Get or Set operations.**
"


I believe Path.GetFullPath( "E:" ) had returned the working directory because it was the active directory on the drive, and all I did was specify a drive letter, not a fully qualified path such as `"E:\"`
`E:` refers to the drive itself, while `E:\` refers to the root directory of the drive. Windows default functionality may be that if path is not fully qualified, assume the 'currently active directory'. which is likely why robocopy AND path.GetFullPath both showcase this same path.

Even when I attempted to sanitize using DirectoryInfo objects, it returned the working directory of the drive instead of the root when submitting `E:` as the object's path.

BUT, based on my testing, I can't reproduce the backslash issue you're referring to. That seems to be working properly, the issue is if the application's based on the drive you are copying the root of, which the solution, as you had said, was to update the  working directory. 

Still debugging my boolean check.

PCAssistSoftware:
> " Although each process can have only one current directory, if the application switches volumes by using the SetCurrentDirectory function, the system remembers the last current path for each volume (drive letter). **This behavior will manifest itself only when specifying a drive letter without a fully qualified path when changing the current directory point of reference to a different volume. This applies to either Get or Set operations.** "

Yes this was definitely the issue I was seeing when just using RoboCopy from command line and as explained with examples here - https://github.com/tjscience/RoboSharp/issues/116#issuecomment-986251066 - e.g. failing when changing back to C: because it was remembering the last current path for the previous drive letter (E:), in that example case E:\Programming\Visual Basic\Apps Developed\RoboCopy GUI\RoboCopy GUI\bin\Debug

But the issue I was commenting on in my previous reply was to do with differences between how VB and C# handle a string containing a single slash which is what explains differences between reported issues with Backup App (in C#) and me testing same thing in VB

RFBomb:
When VB passes a string of `"C:\"` into a C# DLL, C# will interpret that as `"C:\\"`, since C# requires escaping the backslash. It should be done automatically. The conversion back should also be automatic.  Are you saying that is not the case?

PCAssistSoftware:
> When VB passes a string of `"C:\"` into a C# DLL, C# will interpret that as `"C:\\"`, since C# requires escaping the backslash. It should be done automatically. The conversion back should also be automatic. Are you saying that is not the case?

Let's leave this point for now as to be honest I think I am just confusing myself let alone not helping you.  Let's get one issue resolved and then I can go back to that, test and try to explain if still an issue.

RFBomb:
@PCAssistSoftware  try now. 

I have also updated the backup app to work around the issue.

PCAssistSoftware:
@RFBomb 

Tested with various source paths in BackupApp and with my own GUI and that seems to work fine :) - Thank you

Couple of related queries:-

1) In Backup App you have commented it as //Removable Drive Fix  -  why Removable? I have / had this issue for all root drives not just removable ones?  

2) Very odd behaviour here - I am copying from Source: G: to Destination: C:\Users\darre\Desktop\folder2  - and running the app from G:\Debug- with your fixes it works perfectly fine now, but after the copy is complete the folder2 on my desktop is set as hidden - see video clip of this here = https://www.screencast.com/t/nSJhUEZP0

Seems it is setting both hidden AND system attributes to the destination folder. Only happens when source is root of a drive, so perhaps it is copying attributes to it and treating as if it is a drive

RFBomb:
The issue will occur on any drive root if the working directory is on that
drive. Yes. But how often will someone actually use a program on a drive to
copy the entire drive contents elsewhere? Is that really a highly used
use-case? Typically, I would expect that functionality from a program that
installs itself off of a USB or external hard drive, not, for example: the
C drive.

The comment just reflects what I perceived to the most common occurrence of
that situation, feel free to adjust.

Feedback on the new property and method descriptions within RoboCommand.cs
is welcome as well.

As for the hidden folder, mine didn’t hide when I tested it using the
example app. Likely has to do with the selected copy options.

PCAssistSoftware:
> The issue will occur on any drive root if the working directory is on that drive. Yes. But how often will someone actually use a program on a drive to copy the entire drive contents elsewhere? Is that really a highly used use-case? Typically, I would expect that functionality from a program that installs itself off of a USB or external hard drive, not, for example: the C drive. 

Fair point, but I have done it, as my app is not portable and has an installer, so could be installed to C: and then C: is my copy source - as you say rarely but still possible

> The comment just reflects what I perceived to the most common occurrence of that situation, feel free to adjust. 

Okay will think about it, just to ensure makes sense to those coming across it later

> Feedback on the new property and method descriptions within RoboCommand.cs is welcome as well. 

It reads really well to me and definitely explains the issue.  I like how you have done it so it is up to the calling application to set it with `SetWorkingDirectoryToSource`

> As for the hidden folder, mine didn’t hide when I tested it using the example app. Likely has to do with the selected copy options.

Just using defaults set in Backup App and it does it every single time.

Seems it is setting both hidden AND system attributes to the destination folder. Only happens when source is root of a drive, so perhaps it is copying attributes to it and treating as if it is a drive.

PCAssistSoftware:
Thank you for your work on this Robert - much appreciated :)

RFBomb:
Are you using Windows 11? I am using Windows 7 currently. Windows 11 may have introduced an issue that caused that.
I cannot duplicate the destination folder going hidden issue. 


>>The comment just reflects what I perceived to the most common occurrence of that situation, feel free to adjust.

>Okay will think about it, just to ensure makes sense to those coming across it later

When they review the code, if in a designer like VS, hovering will also show the description provided to the method. (I also put it in the call method to ensure that the method worked, since the method itself was in the DLL and not the backup app window.)

PCAssistSoftware:
Yes I am using Windows 11, does same on Windows 10 - will investigate, but suspect it is just to do with copying of attributes.

:)

Will wait to see if any other testers / comments and if okay then merge.

Thanks again.

RFBomb:
I also suspect that the issue you are describing is due to copying attributes over to the destination folder. Since that is not really related to the issue of the working directory, that should go into its own PR, as that is likely going to be a bit more intensive, not only to investigate but to fix. It could also be an issue with robocopy itself on the newer OS, if they changed it for some reason.

PCAssistSoftware:
agreed, if it is specific to my use case or OS then I can just use result of WorkingDirectoryConflictExists to set attributes on destination folder after copy if needed

RFBomb:
> agreed, if it is specific to my use case or OS then I can just use result of WorkingDirectoryConflictExists to set attributes on destination folder after copy if needed

This may be tricky. Once the working directory is changed to a non-conflicting directory, that property will return false. This is by design. Since `get` is a method, it will run every time the property is accessed, ensuring that the most up to date result is returned. I wrote it in such a way to ensure its dynamic, since you never know when the property will be evaluated, or the source or working directory may be changed. (dynamic also requires less ongoing maintenance if other things change)

To do what you suggest,. you will have to essentially:

- Create a new bool in your code to store the result of the property.
- if true 
     - Store the current working directory into a string to be able to restore it (if needed)
     - Store the attributes of the destination folder for use later
     - call `SetWorkingDirectoryToSource`
- Run the Start command
- Evaluate the stored bool a second time
     - If desired, restore original working directory (not likely needed)
     - Do the fix to the top-level folder attributes
     


One last thought about your confusion from before: 
 
>> When VB passes a string of "C:\" into a C# DLL, C# will interpret that as "C:\\", since C# requires escaping the backslash. It should be done automatically. The conversion back should also be automatic. 

I realized your confusion there may be partly my wording. C# is not _interpreting_ `"C:\"` as `"C:\\"`, it is still equivalent to `"C:\"`. But due to syntax, when hovering or checking out the raw value, Visual Studio will display the extra slash (the escape character). 
When writing the value to console, the correct number of slashes are present ( 1 ). BUT if you go to live-edit the value during debugging (or set up a string="X" line), that is when you have to escape them. 

The same thing occurs for other special characters, such as quotes. For example:  `"\"TestQuote\""` evaluates to `"TestQuote"`.
The quote symbol must be escaped here because `"` is treated as a start or end of a string value.
In VB, you might do something like:  `'"' + "TestQuote" + '"'` (which is also valid in C# since `'` denotes a character and not a string). (To get a single `'` as a char though, it would look like this: `'\''`, while as a string it would be: `"'"` due to escaping requirements)
Since the backlash is the escape character, in order to use it within a string or character, you must escape it, which creates the double slash appearance. (Other special escaped characters include NewLine `\n`, Carriage return `\r\n`, tab `\t`, etc)

PCAssistSoftware:
> This may be tricky. Once the working directory is changed to a non-conflicting directory, that property will return false. This is by design...................

Good point, hadn't thought that through! Not a problem as I can just do it based on value of source or similar - will play around with it and make sure it is same issue on other computers and not just something specific to my build environment first.

> One last thought about your confusion from before:
> 
> > > When VB passes a string of "C:" into a C# DLL, C# will interpret that as "C:\", since C# requires escaping the backslash. It should be done automatically. The conversion back should also be automatic......

Yes I understand, think of a lot of it was me confusing myself from results I was getting from RoboSharp source I was using, which as it turned out whilst testing this PR had issues, so once got clean source PR worked and my previous issues didn't exist - so think all down to me.....

JoE-Seeve:
@RFBomb & @PCAssistSoftware 
Last night I carried out a number of tests (and also changes) and found that the problems we had discussed in the last few days (paths, current directory, copying of root directories, etc.) are all due to the path handling of RoboSharp and can be eleminated. Therefore one should not close all points there.

The solution is relatively simple but needs some explanation I guess.

Now I have a working RoboSharp version that now also works with VB.NET. I would like to describe the solution, but I think this is not the right place here, is it?

RFBomb:
Fork the repo, apply your changes, and submit as a pull request for review
and testing.

In that pull request please detail what the problem being solved is as
well, as at this point I’m not sure what the problem was you are referring
to.

A new PR would be best place for discussion, since this PR is not dealing
with VB compatibility.a good name for the PR would be “Improve VB.Net
compatibility”

Thanks

JoE-Seeve:
OK, I'll open a pull request. I won't use the title because it doesn't really have anything to do with VB.NET. @RFBomb, I have now noticed that I have too little knowledge to create a pull request correctly (even the help was not really helpful to me). Could you help me set up a PR that solves problems #116 and #119?

PCAssistSoftware:
@JoE-Seeve - Hi, please explain in more details (wtih examples if possible) for both me and @RFBomb what parts of #116 and #119 are NOT solved by this PR? as from all the testing both me and Robert have done the changes he has made in #119 fix the issue I originally raised in #116

JoE-Seeve:
OK, here are my results and solutions. Two changes are necessary to solve the problems and I will describe each one individually. But they only work if both changes are integrated.

Problem 1.) I tested with robocopy why there are problems with root directories. I did two tests:
Test 1: The test does not work as expected because it references the CurrentDirectory.
grafik

Test 2: Test works, but not with RoboSharp. RoboSharp deletes all "\\" at the end so that reference is always made to the CurrentDirectory.
grafik

Solution for Problem 1: Delete the function that removes trailing slashes (CleanDirectoryPath in ExtensionMethods.cs)
grafik

RFBomb:
Sanitizing by removing trailing directory slash is still required for
non-root paths.

https://superuser.com/questions/1544437/why-does-the-trailing-slash-on-the-target-confuse-robocopy

Better solution may be to check for length == 2 and add the slash back in
in that scenario.

A UNC path may also pose a problem here.

JoE-Seeve:
Problem 2.) Now RoboSharp only works partially and sometimes not (and my VB.NET app no ​​longer works either). What is the cause here now?

Test 1: This test does not work!
grafik

Test 2: This test is now working. If this test works, why does RoboSharp keep adding double quotes around the paths if this doesn't work?
grafik

Solution for Problem 2: According to the RoboCopy documentation, paths with spaces should be quoted. RoboSharp always does this! As a result, the program code has been changed so that it only adds quotation marks if there are spaces in the path (Parse in CopyOptions)
grafik

That's all!

RFBomb:
A while back I had already put a method into the error logging class to
wrap parts only if needed. That can be reused here for more inline code.

Today I will make that change and also see about the trailing slash for
roots and update this PR with a new commit.

JoE-Seeve:
Please test these changes. For me they have eliminated all the problems listed in #116 and #119 (including trailing slashes and also in my VB.NET app).

PCAssistSoftware:
Problem 1 - Test 1 - this now works fine in RoboSharp due to changes made in 119.  In RoboCopy it is a known issue and simply making sure working directory is same as source solves it e.g. CD C:\ in the case of your example

Problem 1 - Test 2 - seems to work fine for me in RoboSharp after changes made in 119.

RFBomb:
I do think that with his proposed change we may be able to work around
having to update the CurrentDirectory when dealing with root dirs on the
same drive.

Per my previous link, we can just do the escaped slash or `“E:\.”` if a
root is detected. No need to change the working dir if that works.

JoE-Seeve:
@PCAssistSoftware Changing the working directory is no longer necessary with the two solutions I have proposed!!!!!

The reason is that my changes no longer delete the slashes at the end (and thus there are no longer any references to the start directory) and root directories are not wrapped in quotation marks. My tests with RoboCopy clearly show that, don't they?

PCAssistSoftware:
> I do think that with his proposed change we may be able to work around having to update the CurrentDirectory when dealing with root dirs on the same drive. Per my previous link, we can just do the escaped slash or `“E:\.”` if a root is detected. No need to change the working dir if that works.

> @PCAssistSoftware Changing the working directory is no longer necessary with the two solutions I have proposed!!!!!

Okay, does it still work if you are running RoboSharp from same drive you have as source, as that was my original issue

Look forward to testing it

RFBomb:
Joe,

I fear that your changes as is may break something though. Quotes are
required if a space exists in the string. Leaving trailing slashes will
escape the end-quote, as you found out. So it must be modified to work I’m
both scenarios, not just root and strings without spaces. (Which is why the
quotes were added originally, which then caused removal of the slashes).

Removing trailing slashes also ensures that a path like `F:\test\\\\` isn’t
sent through.

JoE-Seeve:
@RFBomb "F:\test\\\\" isn't a path. You can check a path with "System.IO.Path.GetInvalidPathChars".

The best way is to check a given path and if the path is invalid, throw an exception.

RFBomb:
`F:\test\` is a path. If some application currently submits an extra \ (
F:\test\\ ) at the end, removing the ‘ remove trailing slash ‘ loop will
break that application. Which is why I wanted to avoid removing it. It’s
sanitizing the input string, better to improve the sanitization process
than remove something that may be useful.

JoE-Seeve:
@RFBomb That's exactly the bug that is built into RoboSharp! If you delete ending slashes then you change "C: \\" (= root directory) to "C:" (= CurrentDirectory). And I think that the users of RoboSharp do not want this!

The problem with simple deletion is when do you know whether the ending slash is necessary and when not?

RFBomb:
You misunderstand and I don’t feel like explaining it again.

JoE-Seeve:
OK, I'm sorry if I didn't understand something here. I apologize.

RFBomb:
@PCAssistSoftware Please re-test this new commit. (it worked perfect during my testing, no WorkingDirectory change required!)

@JoE-Seeve had correctly pointed out that the slash removal was root cause of the issue, but simply removing that loop was not the way to go about fixing it properly in my opinion. 

The loop was required because the path was always wrapped in quotes, even when not needed. So, per his recomendation (and similar to what I had PR'd for the logging options), CopyOptions now has a WrapPath method to sanitize the path for the argument generation.

The loop for removing slashes was adjusted to only remove if length greater than 3 characters, since shortest valid path is 3 characters long `C:\`

Just in case, I also put in a fix for UNC path roots, since those will be longer than 3 characters. A mapped network drive will look like `//server/drivename$/` -- this fix adds the slash back onto here to prevent `//server/drivename$` so that both UNC and Fixed rooted work the same.

PCAssistSoftware:
@RFBomb 

Agreed

Done some initial testing using Backup App and works fine with E:\ and UNC paths.

BUT if I just enter E: in source or destination e.g. with no trailing slash then I get error as below:-

image

Edit: Just tested using my own VB.NET GUI and working fine for all paths tried, and with UNC paths.  And working fine when copying from root source of same drive app is running from :)

Only issue is as above if you were to manually enter a path of just C: , or E: rather than C:\ or E:\

RFBomb:
I think that we should prevent using paths like `E:` as it is in invalid path. As such, it should be sanitized to `E:\`

agree?

PCAssistSoftware:
> I think that we should prevent using paths like `E:` as it is in invalid path. As such, it should be sanitized to `E:\`
> 
> agree?

definitely agree

RFBomb:
By rights, someone could also simply enter in a single letter (or number) here. 

How should we handle that?

As in, a user just typing in `E` or `1` ? 
`E` could be sanitized to `E:\`, but I'm hesitant to do that. In my opinion, anything less than 2 characters ( `E:`) is invalid and should simply return String.empty, while 2 characters for a valid drive root can become `E:\`
 (Really, a single letter or an invalid character should be handled by the caller)

PCAssistSoftware:
That is true, in my GUI I have made it so you can only use FolderBrowserDialog to select source / destination to avoid such issues, but agreed it would be better to handle all invalid input.

Not in front of computer at the moment, but doesn't it already check if it is a valid path? using something like If Directory.Exists or similar? that would get round most invalid input?

I personally tend to handle checks like that in my GUI so nothing invalid ever passed to DLL but good to cover all angles I suppose.

RFBomb:
OK, this next commit should fix that. 
I tested it using `1` and `F`, which both produced "Source Directory does not exist!" fault, while `F:` resulted in a successful copy.

PCAssistSoftware:
That works fine re invalid entries etc

But sorry to say I am still having an issue here ref working directory:-

I have two drives connected - F: and G:
I am running Backup App from Debug folder on the F: drive

image

If I enter F: as source and G: as destination then it works fine and everything is copied to G:

BUT if I enter source as f: (in lowercase rather than capitals) then it only copies working directory of app and not root of F:

PCAssistSoftware:
Below shows content of CommandOptions for both tests above - appears when using lowercase it isn't appending \ to end?

image

RFBomb:
regex adjusted.

PCAssistSoftware:
Perfect - thanks Robert - seems to work with all tests I have run.  Will wait to see if @JoE-Seeve (or others) want to test or has any comments.... and then merge

JoE-Seeve:
As I know "E:" (or "C:", etc.) is a valid path and points to the CurrentDirectory. You can check this in CMD.

@PCAssistSoftware "BUT if I enter source as f: (in lowercase rather than capitals) then it only copies working directory of app and not root of F:" -> This is exact that what I know about this path structure and should be correct.

RFBomb:
This whole discussion started due to  `E:` choosing the incorrect directory due to the app's working directory potentially being on the same drive. This whole endeavor was to prevent that issue from occuring, since many apps likely aren't going to be tossing around their active directory.
Calling apps should specify a proper path. This library should not assume that the current directory is the desired location to copy.  If that is what the calling app wants to do, then they should pass that path in as a string.

The extra sanitization that was introduced was aimed to optimize the sanitization method, but also to ensure that roots are properly handled for both UNC and standard path strings, and to eliminate the issue described in #116

PCAssistSoftware:
> This whole discussion started due to E: choosing the incorrect directory due to the app's working directory potentially being on the same drive. This whole endeavor was to prevent that issue from occuring, since many apps likely aren't going to be tossing around their active directory.

And the endeavour was worth it as the fixes work perfectly to achieve that. 👍

> Calling apps should specify a proper path. This library should not assume that the current directory is the desired location to copy. If that is what the calling app wants to do, then they should pass that path in as a string.

Totally agreed - if you specify f:, F:, F:\ or f:\ then it means root of the F drive and should behave as it now does. 👍 

If you want to copy the working directory of your app then you would enter the full path of that directory e.g. F:\Debug, not just assume it will find it from f: etc - as the results of that even using RoboCopy are just not consistant and change depending on what drive / path you have selected in command window, what last working path was on previous selected drive etc etc - all as covered in previous test results and explained in post by @RFBomb above - https://github.com/tjscience/RoboSharp/pull/119#issuecomment-988094586

PCAssistSoftware:
@RFBomb - unless you have any reasons why not then I will merge this tomorrow as all seems to be working fine here

PCAssistSoftware:
@RFBomb 

> I also suspect that the issue you are describing is due to copying attributes over to the destination folder. Since that is not really related to the issue of the working directory, that should go into its own PR, as that is likely going to be a bit more intensive, not only to investigate but to fix. It could also be an issue with robocopy itself on the newer OS, if they changed it for some reason.

Just for info - the issue of destination being marked as hidden / system I mentioned when source is root and destination is a folder is a known issue which is covered in more detail with solution here https://blog.coeo.com/how-to-prevent-robocopy-from-hiding-your-files and https://stackoverflow.com/questions/6584402/after-robocopy-the-copied-directory-and-files-are-not-visible-on-the-destinatio#:~:text=RoboCopy%20may%20set%20the%20new,switch%20to%20your%20robocopy%20command

Update: can't recreate the problem since latest fixes so perhaps issue was caused by either the original working directory fix we tried, or is resolved because of changes made in this PR relating to paths etc. :)

---

Unless you have any reasons why not then I will merge this pull later today or tomorrow as all seems to be working fine here

RFBomb:
I think the changes we made make sense, and (unless someone is abusing the
current directory bug) it shouldn’t impact anyone.

Issue #122, opened 2021-12-14
Does not work with the newest Version of .net 6.0lts


MarkECrum:
Ok, I am quite new to .net. So if this is not how to post a question, I apologize upfront. I am not what I call an expert programmer by no means but have developed applications in the past.... Just keep that in mind. I Have gone through the code and pretty much understand what is going on with it. However, what I don't understand to a strong degree (because I have not done it yet) is the whole .net targeting and distribution stuff. I do know this.... It shows the dependencies to use lower .net versions. Can this be updated on my end? or What do I have to do to Update it for the latest version?

PCAssistSoftware:
Hi @MarkECrum 

Just created a new WinForms app using .NET 6.0 LTS, manually added reference to RoboSharp.dll (.NET 4.5 version) and it works perfectly fine for me.

Just tested adding RoboSharp via Nuget instead of adding it manually and that also worked fine

PCAssistSoftware:
@tjscience @RFBomb

Related side note after doing quick test for above response - I notice when I add RoboSharp via NuGet to a .NET 6 project that .StartAsync is not available (although it is available when I manually add a reference to the RoboSharp .NET 4.5 DLL rather than using NuGet)

I think we perhaps need to change the preprocessor directives on the below line

https://github.com/tjscience/RoboSharp/blob/b55f5193c558c5405f2e1e470ee2ee9b372b7483/RoboSharp/RoboCommand.cs#L229

from NET45 to NET45_OR_GREATER so it is available on all .NET versions **newer** than 4.5 rather than **just** 4.5? Or am I misunderstanding how it works?

Not sure if we will also need to add preprocessor directives on same line so StartAsync is available for .NET Standard, .NET 5, .NET 6 etc?

Just a thought for discussion

MarkECrum:
Hmmmm, I'll create  a new one mirroring your example with a few switches
and try it again... thanks, I'll let you know.

MarkECrum:
I just tested the recreation adding a reference manually to the dll. It does not work.... or that is, I have to right click and tell it to "Run as Administrator" Funny though, I have local admin rights!???? When I added this all seemed like it worked. Can anyone tell me what that is all about? Besides the /zb or /b should take care of this... should it not?

PCAssistSoftware:
[...]

PR #127, opened 2021-12-21
Create RoboCopyResultsList Object for Processing Multiple RoboCopyResults Objects

RFBomb:
Create a List Object for RoboCopyResults that allows auto-summing of the Statistics Objects contained within the list's Results objects.

- Since this may be used for binding, Implement `INotifyPropertyChanged` on the Statistics and SpeedStatistics objects

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ cmd.OnCommandCompleted += (args) =>
 cmd.Start();
 ```
 
+N.B. The below has been superseded by changes in PR #127 - documentation will be updated shortly to cover all new methods
+
 .AddStatistic
 
 This is useful if you are running multiple RoboCopy tasks as it allows you to add all the statistics to each other to generating overall results
@@ -145,8 +147,6 @@ void copy_OnCommandCompleted(object sender, RoboCommandCompletedEventArgs e)
             }));
         }
 ```
-
-N.B. The below has been superseded by changes in PR #127 - documentation will be updated shortly to cover all new methods
 
 .AverageStatistics
 
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ void backup_OnCommandCompleted(object sender, RoboCommandCompletedEventArgs e)
 }
 ```
 
-Extended Results:
+#### Extended Results:
 
 See below examples on how to access the extended results where you can get total, copied, skipped, mismatch, failed and extra statistics for directories and files and bytes, as well as additional speed info e.g. bytes per sec and megabytes per minute
 
@@ -101,6 +101,72 @@ cmd.OnCommandCompleted += (args) =>
 }
 cmd.Start();
 ```
+
+.AddStatistic
+
+This is useful if you are running multiple RoboCopy tasks as it allows you to add all the statistics to each other to generating overall results
+
+```c#
+RoboSharp.Results.Statistic FileStats = new RoboSharp.Results.Statistic();
+RoboSharp.Results.Statistic DirStats = new RoboSharp.Results.Statistic();
+
+test = new RoboCommand();
+
+// Run first task and add results
+test.CopyOptions.Source = @"C:\SOURCE_1";
+test.CopyOptions.Destination = @"C:\DESTINATION";
+
+RoboSharp.Results.RoboCopyResults results1 = await test.StartAsync();
+
+FileStats.AddStatistic(results1.FilesStatistic);
+DirStats.AddStatistic(results1.DirectoriesStatistic);
+
+// Run second task and add results
+test.CopyOptions.Source = @"C:\SOURCE_2";
+test.CopyOptions.Destination = @"C:\DESTINATION";
+
+RoboSharp.Results.RoboCopyResults results2 = await test.StartAsync();
+
+FileStats.AddStatistic(results2.FilesStatistic);
+DirStats.AddStatistic(results2.DirectoriesStatistic);
+```
+
+You could also use .AddStatistic in the OnCommandCompleted event e.g.
+
+```c#
+void copy_OnCommandCompleted(object sender, RoboCommandCompletedEventArgs e)
+        {
+            this.BeginInvoke((Action)(() =>
+            {
+                // Get robocopy results 
+                RoboSharp.Results.RoboCopyResults AnalysisResults = e.Results;
+
+                FileStats.AddStatistic(AnalysisResults.FilesStatistic);
+            }));
+        }
+```
+
+
+N.B. The below has been superseded by changes in PR #127 - documentation will be updated shortly to cover all new methods
+
+.AverageStatistics
+
+Again if running multiple RoboCopy tasks you can use this to get average results for BytesPerSec and MegaBytesPerMin 
+
+Based on above example
+
+```c#
+// Call Static Method to return new object with the average
+RoboSharp.Results.SpeedStatistic avg = RoboSharp.Results.SpeedStatistic.AverageStatistics(new RoboSharp.Results.SpeedStatistic[] { results1.SpeedStatistic, results2.SpeedStatistic });
+```
+
+or
+
+```c#
+// Result1 will now store the average statistic value. Result 2 can be disposed or or re-used for additional RoboCopy commands.
+results1.AverageStatistic(results2);
+```
+
 =======
 
 # Contributing to RoboSharp
--- a/RoboSharp.BackupApp/MainWindow.xaml
+++ b/RoboSharp.BackupApp/MainWindow.xaml
@@ -202,7 +202,7 @@
                             </Expander>
                         </StackPanel>
                     </ScrollViewer>
-                    
+
                     <Button Content="Backup" HorizontalAlignment="Right" Margin="0,0,10,10" VerticalAlignment="Bottom" Width="75" Click="Button_Click"/>
                 </Grid>
             </TabItem>
@@ -225,11 +225,12 @@
             <TabItem Header="Errors" Name="ErrorsTab">
                 <DataGrid Name="ErrorGrid"></DataGrid>
             </TabItem>
-            <TabItem Header="TabItem" HorizontalAlignment="Left" Height="20" VerticalAlignment="Top" Width="57">
+            <TabItem Header="Job History" HorizontalAlignment="Left" Height="20" VerticalAlignment="Top" Width="57">
                 <Grid x:Name="ResultsListGrid" Background="#FFE5E5E5">
-                    <ListBox x:Name="ListBox_JobResults" HorizontalAlignment="Left" Height="320" Margin="10,10,0,0" VerticalAlignment="Top" Width="241" SelectionChanged="UpdateSelectedItemsLabel"/>
-                    <Label x:Name="lbl_OveralTotals" Content="Label" HorizontalAlignment="Left" Height="123" Margin="256,10,0,0" VerticalAlignment="Top" Width="234"/>
-                    <Label x:Name="lbl_SelectedItemTotals" Content="Label" HorizontalAlignment="Left" Height="123" Margin="256,197,0,0" VerticalAlignment="Top" Width="234"/>
+                    <ListBox x:Name="ListBox_JobResults" HorizontalAlignment="Left" Height="320" Margin="10,54,0,0" VerticalAlignment="Top" Width="241" SelectionChanged="UpdateSelectedItemsLabel"/>
+                    <Label x:Name="lbl_OveralTotals" Content="Total History Results" HorizontalAlignment="Left" Height="123" Margin="256,187,0,0" VerticalAlignment="Top" Width="234"/>
+                    <Label x:Name="lbl_SelectedItemTotals" Content="Selected Job Results" HorizontalAlignment="Left" Height="123" Margin="256,59,0,0" VerticalAlignment="Top" Width="234"/>
+                    <Label x:Name="lbl_OveralTotals_Copy" Content="This tab contains a list of the results from all previous runs during this session." HorizontalAlignment="Left" Height="44" Margin="10,10,0,0" VerticalAlignment="Top" Width="669"/>
                 </Grid>
             </TabItem>
         </TabControl>
--- a/RoboSharp.BackupApp/MainWindow.xaml
+++ b/RoboSharp.BackupApp/MainWindow.xaml
@@ -228,7 +228,7 @@
             <TabItem Header="Job History" HorizontalAlignment="Left" Height="20" VerticalAlignment="Center" Width="94" Margin="-2,0,0,0">
                 <Grid x:Name="ResultsListGrid" Background="#FFE5E5E5">
                     <ListBox x:Name="ListBox_JobResults" HorizontalAlignment="Left" Height="320" Margin="10,54,0,0" VerticalAlignment="Top" Width="321" SelectionChanged="UpdateSelectedItemsLabel"/>
-                    <Label x:Name="lbl_OverallTotals" Content="Total History Results" HorizontalAlignment="Left" Height="123" Margin="336,251,0,0" VerticalAlignment="Top" Width="367"/>
+                    <Label x:Name="lbl_OverallTotals" Content="Total History Results" HorizontalAlignment="Left" Height="179" Margin="336,251,0,0" VerticalAlignment="Top" Width="367"/>
                     <Label x:Name="lbl_SelectedItemTotals" Content="Selected Job Results" HorizontalAlignment="Left" Height="192" Margin="336,54,0,0" VerticalAlignment="Top" Width="367"/>
                     <Label x:Name="lbl_JobHistoryHeader" Content="This tab contains a list of the results from all previous runs during this session." HorizontalAlignment="Left" Height="44" Margin="10,10,0,0" VerticalAlignment="Top" Width="669"/>
                 </Grid>
--- a/RoboSharp.BackupApp/MainWindow.xaml
+++ b/RoboSharp.BackupApp/MainWindow.xaml
@@ -225,12 +225,12 @@
             <TabItem Header="Errors" Name="ErrorsTab">
                 <DataGrid Name="ErrorGrid"></DataGrid>
             </TabItem>
-            <TabItem Header="Job History" HorizontalAlignment="Left" Height="20" VerticalAlignment="Top" Width="57">
+            <TabItem Header="Job History" HorizontalAlignment="Left" Height="20" VerticalAlignment="Top" Width="94" Margin="-2,-2,-35,0">
                 <Grid x:Name="ResultsListGrid" Background="#FFE5E5E5">
                     <ListBox x:Name="ListBox_JobResults" HorizontalAlignment="Left" Height="320" Margin="10,54,0,0" VerticalAlignment="Top" Width="241" SelectionChanged="UpdateSelectedItemsLabel"/>
-                    <Label x:Name="lbl_OveralTotals" Content="Total History Results" HorizontalAlignment="Left" Height="123" Margin="256,187,0,0" VerticalAlignment="Top" Width="234"/>
-                    <Label x:Name="lbl_SelectedItemTotals" Content="Selected Job Results" HorizontalAlignment="Left" Height="123" Margin="256,59,0,0" VerticalAlignment="Top" Width="234"/>
-                    <Label x:Name="lbl_OveralTotals_Copy" Content="This tab contains a list of the results from all previous runs during this session." HorizontalAlignment="Left" Height="44" Margin="10,10,0,0" VerticalAlignment="Top" Width="669"/>
+                    <Label x:Name="lbl_OverallTotals" Content="Total History Results" HorizontalAlignment="Left" Height="123" Margin="256,187,0,0" VerticalAlignment="Top" Width="367"/>
+                    <Label x:Name="lbl_SelectedItemTotals" Content="Selected Job Results" HorizontalAlignment="Left" Height="123" Margin="256,59,0,0" VerticalAlignment="Top" Width="367"/>
+                    <Label x:Name="lbl_JobHistoryHeader" Content="This tab contains a list of the results from all previous runs during this session." HorizontalAlignment="Left" Height="44" Margin="10,10,0,0" VerticalAlignment="Top" Width="669"/>
                 </Grid>
             </TabItem>
         </TabControl>
--- a/RoboSharp.BackupApp/MainWindow.xaml
+++ b/RoboSharp.BackupApp/MainWindow.xaml
@@ -225,11 +225,11 @@
             <TabItem Header="Errors" Name="ErrorsTab">
                 <DataGrid Name="ErrorGrid"></DataGrid>
             </TabItem>
-            <TabItem Header="Job History" HorizontalAlignment="Left" Height="20" VerticalAlignment="Top" Width="94" Margin="-2,-2,-35,0">
+            <TabItem Header="Job History" HorizontalAlignment="Left" Height="20" VerticalAlignment="Center" Width="94" Margin="-2,0,0,0">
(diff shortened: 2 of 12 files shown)


RFBomb:
@PCAssistSoftware  - Try this out! 
Instead of calling the adding/averaging methods yourself, create the RoboCopyResultsList() object, which is simply a List<RoboCopyResults> object with custom properties for auto-summing. 

Since I tried to go for optimization (since this is a library), I used Lazy objects to avoid doing the math until the result is requested for the first time, then update that statistic object once it has been created initially. This means you to inspect the value, you will likely have to request it first. 

Here is some mockup vba (I think):

dim RoboCopyResultsList ResultsList as RoboCopyResultsList
dim Statistic FileResults as Statistic 
ResultsList = new RoboCopyResultsList()
FileResults = ResultsList.Total_FilesStatistic

Then you should be able to inspect FileResults itself. Note: Doing math against FileResults WILL screw up the value, since the value is not recalculated 100% every time the list is changed. Instead, it simply adds the new result object to the existing statistic (instead of resetting the values to zero and looping through all result obiects every time a new result is added)

A Recalculate method can easily be added to avoid/fix this issue. Exit Status is also no implemented yet.

PCAssistSoftware:
@RFBomb - now downloading changes with GitHub Desktop

Can you provide me some examples in C# so I can give it a try

RFBomb:
Give me a half hour or so, I’ll put out another commit with it semi-baked
into the backup app

PCAssistSoftware:
N.B. I had posted above response asking for examples before realizing you had already posted one above :) sorry

Amended your mockup VB code as below and it seems to work really fine for me :)

```
Dim ResultsList As Results.RoboCopyResultsList
Dim FileResults As Results.Statistic
ResultsList = New Results.RoboCopyResultsList
FileResults = ResultsList.Total_FilesStatistic
```

With 

`ResultsList.Add(jobresults) `

in my 

`For Each Job In RoboSharpJobsList`

loop

Thanks you Robert - much appreciated

Just out of curiosity why do we have both Extras and ExtrasField, Copied and CopiedField etc - not a problem, just wondered

image

PCAssistSoftware:
>  Exit Status is also no implemented yet.

Bit more awkward for Exit Status, as each 'job' could have a different status which we would need to know e.g. one job may work and another may fail because files already copied

For now I am storing the .Status with each of my "jobs" so I can easily refer to them individually for each job as needed later e.g.

```
For Each Job In RoboSharpJobsList

        Dim jobresults As RoboSharp.Results.RoboCopyResults = Await Job.Command.StartAsync()
		
        ' add exitcode
        Job.ExitCode = jobresults.Status

Next
```

RFBomb:
I have updated the BackupApp and I have updated the ResultsList.
This is a breaking change!

I had noticed that when a new item was added to the list, it was not automatically updating the Listbox's contents. After some research, I found that that was because I was firing a custom event instead of using INotifyCollectionChanged. 

My `ListWithEvents<>` object I created was a rudimentary form of `ObservableCollection<>`.  I prefered the List<> functionality over ObservableCollection though, since it has more methods to add/remove. As such, I have removed the custom event and event args I created in favor of the standard INotifyCollectionChanged event + args, and implemented them accordingly.

I have also renamed the properties to match what everyone is used to based on the current Results object.

The ExitStatus objects for each job are available within the list object if you need them, the one this produces should simply be an overall summary. That way, for example, you can check the overall status before deciding to loop through each one.

RFBomb:
After doing more testing in the backup app, I see the issue with ExitCodes. Fixing now.

RFBomb:
ExitStatus now appears to be combining properly. (I wasn't using the correct operator to combine it)
It should provide an accurate summary now. 

Better yet, we may want to produce a new status object to represent a combined status, instead of stuffing the results and bools into the existing one, simply for clarity of use.

PCAssistSoftware:
Fantastic work as always - thank you Robert - very really appreciated.

Done some quick tests here and all seems to be working as expected.  Will test in more depth tomorrow in my own app and then if okay will merge it.

Only comments I would make are strictly cosmetic ones:-

1. Might be worth widening Job History TabItem header to be wider so text fits
2. Make lbl_SelectedItemTotals and lbl_OveralTotals wider so content fits better
3. Rename lbl_OveralTotals_Copy to something like lbl_JobHistoryIntro
4. Correct spelling (add missing l) so lbl_OveralTotals becomes lbl_OverallTotals

I can make these changes directly in code if easier

RFBomb:
I whipped that page together quickly, I hadn't worked with WPF prior to that, so I missed it lol.
Those changes made this next commit.

PCAssistSoftware:
> I whipped that page together quickly, I hadn't worked with WPF prior to that, so I missed it lol. Those changes made this next commit.

Thanks, no me neither, usually WinForms rather than WPF so took a moment to work it out

PCAssistSoftware:
> Better yet, we may want to produce a new status object to represent a combined status, instead of stuffing the results and bools into the existing one, simply for clarity of use.

Personally I think Exit Status needs some thought, because in my eyes there is not really any point in combining them (other than perhaps for an overall result) because you would need to know 'which status' applied to 'which job' for them to be any use.

e.g. if I ran three jobs and job1 and job2 worked fine but job3 failed with say exit code 8 or 16 then I would need to investigate further, so would need to look at exit code for each job in turn rather than the combined ones to see which job the failed exit code refers too - does that make sense?

That is why I was for now (as above) just associating the Exit Status with the just completed job in my "job list", so at the end I can see as below which code belongs to which job name or job ID:-

image

Also I am storing the whole Status returned so can also see HasErrors, HasWarnings, Successful.

And in my case can see the original RoboCommand so can see what source / destination caused the error.

I have also been playing with LogLines and adding those to my job list so I can then create a log at the end for each job if needed.

PCAssistSoftware:
P.S. The problem with LogLines is it can get really large with some jobs and I have had it crash RoboSharp in the past because of that, especially if migrating large volumes of data across servers etc

RFBomb:
So, I was thinking during this we may want to store the actual command generated into the ExitStatus object for future reference here. 

Just the Source/Destination, and the switches themselves in the form of a string. My idea behind this list is you could already create a list of the results if you wanted to, but you had to parse this manually. Now there is a list object to do that automatically. The exit status serves as an overall summary, but since its a list you can still access all the result objects within the list individually. 

Source Destination I think make sense to include int he result object. OR, we can create a reference to the robocommand that spawned the result. (But if the command object is re-used for multiple runs, you lose your reference to original parameters).

PCAssistSoftware:
> So, I was thinking during this we may want to store the actual command generated into the ExitStatus object for future reference here.
> 
> Just the Source/Destination, and the switches themselves in the form of a string. My idea behind this list is you could already create a list of the results if you wanted to, but you had to parse this manually. Now there is a list object to do that automatically. The exit status serves as an overall summary, but since its a list you can still access all the result objects within the list individually.
> 
> Source Destination I think make sense to include int he result object. OR, we can create a reference to the robocommand that spawned the result. (But if the command object is re-used for multiple runs, you lose your reference to original parameters).

Yes I can see the logic of that and fits in with what I was saying above about associating exit codes with particular jobs for diagnosing failed jobs etc

PCAssistSoftware:
MainWindow.xaml line 228

Change

`   <TabItem Header="Job History" HorizontalAlignment="Left" Height="20" VerticalAlignment="Top" Width="94" Margin="-2,-2,-35,0">`

to

`    <TabItem Header="Job History" HorizontalAlignment="Left" Height="20" VerticalAlignment="Center" Width="94" Margin="-2,0,0,0">`

To move tab back into position

image

RFBomb:
Man, WPF is awkward. That I used the graphical designer to extend it sideway, that shouldn't have even happened.

Pushing this change and adding source/destination to results. Check it out.

PCAssistSoftware:
> Man, WPF is awkward. That I used the graphical designer to extend it sideway, that shouldn't have even happened.
> 
> Pushing this change and adding source/destination to results. Check it out.

Agreed, I really hate working with it, so fiddly doing simple things - usually end up changing things in code rather than designer as easier

RFBomb:
FYI: I didn't add the 'CommandOptions' to the displayed item for the Selected Job because it was only displaying the source. 
I believe this has to do with how WPF is processing the string when applying it to the label, since the parameters had contained `"` around the path. 

That said, the CommandOptions value itself had the expected value when inspecting it using the debugger.

PCAssistSoftware:
> Pushing this change and adding source/destination to results. Check it out.

source and destination showing fine.

> FYI: I didn't add the 'CommandOptions' to the displayed item for the Selected Job because it was only displaying the source. I believe this has to do with how WPF is processing the string when applying it to the label, since the parameters had contained `"` around the path.
> 
> That said, the CommandOptions value itself had the expected value when inspecting it using the debugger.

That is odd, as you say when you look at it in debugger it shows full command, and I use that value to populate a text field okay e.g. textbox1.text = copy.CommandOptions.ToString

RFBomb:
It may have to do with the label not performing text wrapping as well, so the unwrapped portion simply gets cut off, and that happened to be right at the length of my path. I didn't see any option for Text Wrapping.
The label also requires using 'Label.Content', not 'Label.Text' as winforms does.

RFBomb:
Anyway, I think I'm satisfied with this object. It should be fine enough to use I think. 
I wrote it in such a way that I think it should be able to be used in a future 'Jobs' style object that allows queuing multiple RoboCommands,  and as each finishes its results can be added to this list object. 

This way, if the Job object had a method to run all the commands as List-Only, all the results could be added to a list. Then all the copy operations results could be added to a different list. And the list would report the totals instead of devs having to manually combine the results themselves.

RFBomb:
@PCAssistSoftware 
> Just out of curiosity why do we have both Extras and ExtrasField, Copied and CopiedField etc - not a problem, just wondered

The field stores the value that the property interacts with. The field is private, so you can only see it when debugging the library or inspecting it The Property is what is publicly available. This setup allows implementing `INotifyPropertyChanged`

PCAssistSoftware:
> Anyway, I think I'm satisfied with this object. It should be fine enough to use I think. I wrote it in such a way that I think it should be able to be used in a future 'Jobs' style object that allows queuing multiple RoboCommands, and as each finishes its results can be added to this list object.

Agreed, great work - will make working with results from multiple jobs so much easier from now on :)
 
> This way, if the Job object had a method to run all the commands as List-Only, all the results could be added to a list. Then all the copy operations results could be added to a different list. And the list would report the totals instead of devs having to manually combine the results themselves.

Good idea - hadn't thought of that, will play with that idea tomorrow

Thanks again for all your work - much appreciated.  Will test more tomorrow and then merge changes

RFBomb:
Final note: 

a button should be added to be able to remove a selected item from the list box (by removing it from the new list object) to verify that the remove from list and subtract from statistic methods are working properly. 

Removing from the list should cause full recalculation on both speed and status. (All other objects are statistic objects, so they perform simple math here)

Also, I forgot to include speed in the summary labels. So that should be added in.

PCAssistSoftware:
> Final note:
> 
> a button should be added to be able to remove a selected item from the list box (by removing it from the new list object) to verify that the remove from list and subtract from statistic methods are working properly.
> 
> Removing from the list should cause full recalculation on both speed and status. (All other objects are statistic objects, so they perform simple math here)
> 
> Also, I forgot to include speed in the summary labels. So that should be added in.

I'll try and have a good at adding the button etc tomorrow, not great at C# but happy to try

RFBomb:
The button event just needs to call the Remove(item) method of the list,
and the speed label thing is adding it to the Text string in the label
update methods

PCAssistSoftware:
Okay started with adding speed to the text string

Below works fine for UpdateOverallLabel

```
 private void UpdateOverallLabel(object sender, EventArgs e)
        {
            string NL = Environment.NewLine;
            lbl_OverallTotals.Content = $"Job History:" +
                $"{NL}Total Directories: {JobResults.DirectoriesStatistic.Total}" +
                $"{NL}Total Files: {JobResults.FilesStatistic.Total}" +
                $"{NL}Total Size (bytes): {JobResults.BytesStatistic.Total}" +
                $"{NL}Speed (Bytes/Second): {JobResults.SpeedStatistic.BytesPerSec}" +
                $"{NL}Speed (MB/Min): {JobResults.SpeedStatistic.MegaBytesPerMin}" +
                $"{NL}Any Jobs Cancelled: {(JobResults.Status.WasCancelled ? "YES" : "NO")}" +
                $"{NL}{JobResults.Status.ToString()}";
        }
```


But for UpdateSelectedItemsLabel you can't do below as if the selected job didn't have anything to copy then you get a System.NullReferenceException, so not quite so simple as adding it to the string, we will need to check for null / nothing

```
  private void UpdateSelectedItemsLabel(object sender, SelectionChangedEventArgs e)
        {
            Results.RoboCopyResults result = (Results.RoboCopyResults)this.ListBox_JobResults.SelectedItem;
            string NL = Environment.NewLine;
            lbl_SelectedItemTotals.Content = $"Selected Job:" +
                $"{NL}Source: {result.Source}" +
                $"{NL}Destination: {result.Destination}" +
                $"{NL}Total Directories: {result.DirectoriesStatistic.Total}" +
                $"{NL}Total Files: {result.FilesStatistic.Total}" +
                $"{NL}Total Size (bytes): {result.BytesStatistic.Total}" + 
                $"{NL}Speed (Bytes/Second): {result.SpeedStatistic.BytesPerSec}" +
                $"{NL}Speed (MB/Min): {result.SpeedStatistic.MegaBytesPerMin}" +
                $"{NL}{result.Status.ToString()}";
        }
```

PCAssistSoftware:
Thought the below would work e.g. checking if result.SpeedStatistic is null before showing resuts but for some reason it doesn't, could be my understanding of C# compared to VB - as in VB I would use If result.SpeedStatistic IsNot Nothing

 ```
 private void UpdateSelectedItemsLabel(object sender, SelectionChangedEventArgs e)
        {
            Results.RoboCopyResults result = (Results.RoboCopyResults)this.ListBox_JobResults.SelectedItem;
            string NL = Environment.NewLine;
            if (result.SpeedStatistic != null)
                lbl_SelectedItemTotals.Content = $"Selected Job:" +
                $"{NL}Source: {result.Source}" +
                $"{NL}Destination: {result.Destination}" +
                $"{NL}Total Directories: {result.DirectoriesStatistic.Total}" +
                $"{NL}Total Files: {result.FilesStatistic.Total}" +
                $"{NL}Total Size (bytes): {result.BytesStatistic.Total}" + 
                $"{NL}Speed (Bytes/Second): {result.SpeedStatistic.BytesPerSec}" +
                $"{NL}Speed (MB/Min): {result.SpeedStatistic.MegaBytesPerMin}" +
                $"{NL}{result.Status.ToString()}";
            else
                lbl_SelectedItemTotals.Content = $"Selected Job:" +
                $"{NL}Source: {result.Source}" +
                $"{NL}Destination: {result.Destination}" +
                $"{NL}Total Directories: {result.DirectoriesStatistic.Total}" +
                $"{NL}Total Files: {result.FilesStatistic.Total}" +
                $"{NL}Total Size (bytes): {result.BytesStatistic.Total}" +
                $"{NL}Speed (Bytes/Second): n/a" +
                $"{NL}Speed (MB/Min): n/a" +
                $"{NL}{result.Status.ToString()}";
        }
```

**Comment by RFBomb** [comment]
[ Moved into comment below since e-mail replies don’t support markdown]

**Comment by RFBomb** [comment]
Use this instead:

`$"{NL}Speed (Bytes/Second): {result?.SpeedStatistic?.BytesPerSec ?? 0}" + `

The `?.` before the property request is an in-line ` != null ` check. The
`??` afterward sets the what to do in case it is null.

No need for the if statement. If you were to wrap with an if though, it
would be:

```
String X;
if ( condition )
 {  X = VALUE; }
else
{ X = VALUE2; }

```

 While an in-line set value would be :

` string X = ( condition ) ? VALUE : VALUE2; `

The `?` is a true/false check, and `:` sets the else statement.

Combing those to a possible null bool would be:

```
Object X = new Object();
String Y = ( X?.BooleanProperty ?? false ) ? “true” : “false”;

```
So if the object or the value of `BooleanProperty` is null, the statement
returns the string `false` based on the null check statement. If  `
BooleanProperty`
Is true/false, then it reacts according to the value of `BooleanProperty`

RFBomb:
It’s probably worth having the null checks on the other properties in that method as well, I’m case cancellation occurs before the value is generated (robocopy running with multiple threads scenario may produce this)

PCAssistSoftware:
@RFBomb - thanks, that worked fine and I have pushed changed

Query re calculation of average speeds though

Just run 3 jobs, speed results below, but average is not what I would expect

Speed (Bytes/Second): 520.957
Speed (MB/Min): 29.809

Speed (Bytes/Second): 0
Speed (MB/Min): 0

Speed (Bytes/Second): 293.600
Speed (MB/Min): 16.799

Averages shown

Speed (Bytes/Second): 277.03925
Speed (MB/Min): 15.85175

EDIT: will add null checks to rest of properties shortly :)

PCAssistSoftware:
P.S. Thanks for full explanation on null check in C# - very useful for future so added to my notes :)

RFBomb:
It’s because I averaged job 1 & 2, then averaged result & 3. 

I’ll redo the speed statistic today/tomorrow to work better.

PCAssistSoftware:
@RFBomb - remove button added, seems to work fine for me, let me know if you think okay or needs any changes

RFBomb:
I’ll check it out later today

My plan for speed is to derive the speed stat object to a new one to
implement the combine logic, similar to what I did for the status derived
object.

PCAssistSoftware:
That makes sense

RFBomb:
I have created the new AverageSpeedStatistic object, and worked in the functionality to perform the average properly (I believe). 

Note; If no copy operations are performed (or process is cancelled / exit prematurely), then the results builder returns a null SpeedStatistics object apparently. Since no transfer occured, these are ignored. This means the average method will be the average of all jobs that have atleast 1 copy operation performed. 

If an AverageSpeedStatistic object is added to another one, then it works with the underlying values instead of the public values in order to prevent taking an average of two averages. This is documented in the xml comments. 

The internal methods allow for overriding this functionality, but that is not publicly exposed. I wrote in the ability to override it just in case it was needed later on for some reason. 

Since the AverageSpeedStatistics object is derived from a SpeedStatistic, it can be passed around as such, since the publicly exposed properties are identical.

RFBomb:
Should we round the results of the average?

image

PCAssistSoftware:
> I have created the new AverageSpeedStatistic object, and worked in the functionality to perform the average properly (I believe).

Excellent, I will test it later this evening or if not tomorrow and let you know how I get on :)

> Note; If no copy operations are performed (or process is cancelled / exit prematurely), then the results builder returns a null SpeedStatistics object apparently. Since no transfer occured, these are ignored. This means the average method will be the average of all jobs that have atleast 1 copy operation performed.

That makes perfect sense and should work well

> If an AverageSpeedStatistic object is added to another one, then it works with the underlying values instead of the public values in order to prevent taking an average of two averages. This is documented in the xml comments.
> 
> The internal methods allow for overriding this functionality, but that is not publicly exposed. I wrote in the ability to override it just in case it was needed later on for some reason.
> 
> Since the AverageSpeedStatistics object is derived from a SpeedStatistic, it can be passed around as such, since the publicly exposed properties are identical.

Excellent

PCAssistSoftware:
> Should we round the results of the average?
> 
> image

Yes definitely, perhaps 2 decimal places? what do you think?

PCAssistSoftware:
Just amended Remove_Selected_Click event so that if you remove all results then it clears the content of lbl_OverallTotals and lbl_SelectedItemTotals, looks cleaner like that if not old information still show even when ListBox_JobResults is empty

RFBomb:
OK, I think this satisfies this object.

PCAssistSoftware:
> OK, I think this satisfies this object.

Excellent, thanks Robert - been testing on/off all day and seems to work perfectly and is definitely a great addition going forward for all future methods of running multiple 'jobs'.

I will finish testing in my app shortly and then merge the changes.

Thanks again for all your help, and best wishes to you for the festive period :)

RFBomb:
Thanks you too. 

Since I have an idea how to do the multiple jobs, I'm going to start on that as well. I'll start a new branch for that based on this branch. Since my application already had methods for dealing with a list<RoboCommand>, I'm going to just methodize what I already created and start a new PR thread

PCAssistSoftware:
> Thanks you too.
> 
> Since I have an idea how to do the multiple jobs, I'm going to start on that as well. I'll start a new branch for that based on this branch. Since my application already had methods for dealing with a list, I'm going to just methodize what I already created and start a new PR thread

Fantastic, look forward to seeing more

PCAssistSoftware:
I need to update readme.md to cover new methods and changes re old methods

I see the original .AverageStatistics is now .AverageSpeedStatistic.GetAverage

I have copied existing code from documentation below, can you tell me how the changes affect that so I can correct documentation at same time as I merge later - thanks

---

.AverageStatistics

Again if running multiple RoboCopy tasks you can use this to get average results for BytesPerSec and MegaBytesPerMin

Based on above example

// Call Static Method to return new object with the average
RoboSharp.Results.SpeedStatistic avg = RoboSharp.Results.SpeedStatistic.AverageStatistics(new RoboSharp.Results.SpeedStatistic[] { results1.SpeedStatistic, results2.SpeedStatistic });
or

// Result1 will now store the average statistic value. Result 2 can be disposed or or re-used for additional RoboCopy commands.
results1.AverageStatistic(results2);
=======

RFBomb:
The methods that were removed from the objects can be added back in if you want, and we can either restore their old functionality or just make them empty methods. Either way, they should be marked with an `[Obsolete]` tag.

Or, redirect them to the new methods. 
Since my new objects are all derived, it shouldn't break old code

But personally, I think that this clutters up the library quite a bit. Were those methods ever released in a nuget package? If they were, I don't think we have a 'best-practice' choice but to put them back in. If they weren't its debatable.

PCAssistSoftware:
They were never released in a NuGet package :) 

So I agree that it would clutter up the library and perhaps even confuse new users of RoboSharp.

So if agreed I will add new documentation to cover the new methods and get rid of the existing section copied above re .AverageStatistics

RFBomb:
A note in the documentation regarding it makes sense then. I would list out the objects/methods affected then refer to this PR for the changes if they want more info

RFBomb:
Whats the reason behind StartAsync being unavailable except for Net45?

PCAssistSoftware:
> Whats the reason behind StartAsync being unavailable except for Net45?

That should be resolved now in a previous commit - it was previously set for NET45 but it came up in discussion in #122 as I noticed it not working in .NET 6.0 so I amended preprocessor directive so should be available in all versions other than Net40

`#if NET45 || NETSTANDARD2_0 || NETSTANDARD2_1 || NETCOREAPP3_1`

RFBomb:
The GetAwaiter method is missing. Interesting (and inconvenient)
I'll program around that then...

PCAssistSoftware:
> The GetAwaiter method is missing. Interesting (and inconvenient) I'll program around that then...

Okay, not an expert on Async/Await, had just noticed code not updated to allow it for newer .NET versions hence change - hope that was okay, seems to work fine in .NET 6

PCAssistSoftware:
Several hours this PM spent testing all changes, all working as expected thanks to hard work of @RFBomb - now merged

PCAssistSoftware:
@tjscience  

Any chance of publishing an updated NuGet release when you get a moment? To encompass this fantastic PR by @RFBomb  and all the other recent fixes / pulls made which includes fixes for the NuGet documentation issue, enabling StartAsync support in newer .NET versions and so much more.  Thanks :)

PR #129, opened 2021-12-30
Count Files and Dirs in case of Cancellation

RFBomb:
Will provide baseline count for Files and Directories even if job was cancelled.

Since i had my hands in the `RoboCommand.process_OutputDataReceived` method yesterday, I got a good understanding what it does, and knew that tracking the counts could be relatively easy. This PR includes tracking counts to report totals in case of a cancellation.

The fix for the `RoboCopyResults.ToString()` method is here in case that's all that is deemed accepted, I just wanted to mock up the results tracking for your review in addition to it.

```
public override string ToString()
        {
            string str = $"ExitCode: {Status.ExitCode}, Directories: {DirectoriesStatistic?.Total.ToString() ?? "Unknown"}, Files: {FilesStatistic?.Total.ToString() ?? "Unknown"}, Bytes: {BytesStatistic?.Total.ToString() ?? "Unknown"}";

            if (SpeedStatistic != null)
            {
                str += $", Speed: {SpeedStatistic.BytesPerSec} Bytes/sec";
            }

            return str;
        }
```

### DIFF
--- a/RoboSharp/ProcessedFileInfo.cs
+++ b/RoboSharp/ProcessedFileInfo.cs
@@ -5,7 +5,7 @@
     /// </summary>
     public enum FileClassType
     {
-        /// <summary>Details about a FOLDER</summary>
+        /// <summary>Details about a Directory</summary>
         NewDir,
         /// <summary>Details about a FILE</summary>
         File,
@@ -18,10 +18,17 @@
     {
         /// <summary>Description of the item as reported by RoboCopy</summary>
         public string FileClass { get; set; }
+        
         /// <inheritdoc cref="RoboSharp.FileClassType"/>
         public FileClassType FileClassType { get; set; }
-        /// <summary>File Size</summary>
+
+        /// <summary>
+        /// File -> File Size <br/>
+        /// Directory -> Number files in folder -> Can be negative if PURGE is used <br/>
+        /// SystemMessage -> Should be 0
+        /// </summary>
         public long Size { get; set; }
+
         /// <summary>Folder or File Name / Message Text</summary>
         public string Name { get; set; }
     }
--- a/RoboSharp/Results/ResultsBuilder.cs
+++ b/RoboSharp/Results/ResultsBuilder.cs
@@ -10,8 +10,13 @@ namespace RoboSharp.Results
     /// </summary>
     internal class ResultsBuilder
     {
-         
+        private ResultsBuilder() { }
+
+        internal ResultsBuilder(RoboSharpConfiguration config) { Config = config; }
+
         private readonly List<string> outputLines = new List<string>();
+
+        private RoboSharpConfiguration Config { get; }
 
         #region < Command Options Properties >
 
@@ -60,9 +65,9 @@ namespace RoboSharp.Results
         {
             TotalDirs++;
             CurrentDir = currentDir;
-            if (currentDir.FileClass.Contains("Existing")) { /* No Action */ }
-            else if (currentDir.FileClass == "New Dir") { if (CopyOperation) TotalDirs_Copied++; }
-            else if (currentDir.FileClass.Contains("EXTRA")) TotalDirs_Extras++;
+            if (currentDir.FileClass == Config.LogParsing_ExistingDir) { /* No Action */ }
+            else if (currentDir.FileClass == Config.LogParsing_NewDir) { if (CopyOperation) TotalDirs_Copied++; }
+            else if (currentDir.FileClass == Config.LogParsing_ExtraDir) TotalDirs_Extras++;
             else
             {
                 
@@ -84,21 +89,21 @@ namespace RoboSharp.Results
             CopyOpStarted = false;
 
             // EXTRA FILES
-            if (currentFile.FileClass.ToLower().Contains("extra"))
+            if (currentFile.FileClass == Config.LogParsing_ExtraFile)
             {
                 TotalFiles_Extras++;
                 TotalBytes_Extra += CurrentFile.Size;
             }
 
             //MisMatch
-            else if (currentFile.FileClass.ToLower().Contains("mismatch"))
+            else if (currentFile.FileClass == Config.LogParsing_MismatchFile)
             {
                 TotalFiles_Mismatch++;
                 TotalBytes_MisMatch += CurrentFile.Size;
             }
 
             //Failed Files
-            else if (currentFile.FileClass.ToLower().Contains("fail"))
+            else if (currentFile.FileClass == Config.LogParsing_FailedFile)
             {
                 TotalFiles_Failed++;
                 TotalBytes_Failed += currentFile.Size;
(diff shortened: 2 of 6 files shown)


PCAssistSoftware:
I like that, and it works well for me, and I think should definitly be merged as at least gives you some useful output / results upon cancellation rather than nothing at all.  If no further changes then happy to merge it for you?

RFBomb:
Yea I did a quick few tests with it, and cancelled it mid-copy and it
reported fine (225 total, 224 copied)
I could probably get a bit fancier to get more results. Let me. Take a
crack, wanted approval prior to tackling that

PCAssistSoftware:
Okay, sounds good to me, perhaps we can get "some" byte stats as well for cancelled?

RFBomb:
getting byte stats may be more complicated, since we have to track actual file progress here. but should be doable

PCAssistSoftware:
> getting byte stats may be more complicated, since we have to track actual file progress here. but should be doable

Not to worry if it will confuse or complicate matters - just a thought.  All of this will be so useful, especially when running mulitple jobs :)

RFBomb:
Ok, feel free to test and review.
It got a little complicated, but based on this it should also be able to provide some stats even if NJS is used.

PCAssistSoftware:
Great job, initial testing here shows it to be working fine, will run a few more tests shortly

PCAssistSoftware:
One query

Just testing it with NJS set to true

If I let job run to completion then it shows Files: 5  Directories: **0**

If I cancel it after 2 files then it shows Files: 2   Directories: 1

With NJS set to false and running job to completion it then shows Directories: **1**

RFBomb:
My results:

With Summary:
image

NJS:
image

Now, a second attempt with all files already in place:

image

NJS:
image

BUT, if you look at the actual log, the only thing in there is a single directory, so thats all we can record in our estimate.

image

PCAssistSoftware:
The last commit (Fix GetStatisticLines adding paths) seems to have fixed the missing directory value e.g. was showing 0 rather than 1

Same results as you above - so with NJS set to false and files already existing in destination it shows 5 files copied even though they were not copied? surely it should be 0 or is that not possible to do? Total 5 - Skipped 5 = Copied 0

RFBomb:
how did you get 5 copied? My log didn't show any skipped files, so I couldn't do anything

PCAssistSoftware:
So destination already contains the 5 files

image

Running normally e.g. eith NJS at default of false shows 5 files copied when it should surely be 0 as they were not copied as already existed.  Also shows the bytes as 2684354560 when should be 0.

image

Running with NJS set to true shows correct value of 0 files copied and bytes as 0

image

RFBomb:
Thats the Total # of files, not the total # of files copied.
If you actually inspect the ResultsBuilder object during `BuildResults`, it shows up as 5 skipped, 0 copied. 
Also, turning on Verbose log will show 'same' in the log, and an acurrate count can be gotten in the estimate from that. (it will show 5 total since they are now input into the log instead of skipping the log due to skipping the file)

PCAssistSoftware:
Okay, should have known it was total rather than total copied showing in summary.    

EDIT: now added breakpoint so I can check log - and yes all appears fine - my mistake

So if using NJS set to true (which to be honest I am never likely to ever do as the content is useful for me when creating custom log files at the end) then I assume we are obtaining total copied rather than total which is therefore correct?

PCAssistSoftware:
Just run some more tests (with NJS back to default of false) with multiple large files and cancelling at various stages of the operation and results look pefect to me.  Think I have tested all combinations I can think of.

My only confusion above was with NJS set to true (which as I said I wouldn't do anyway) and running to completion as to why it was showing 0 rather than 5 in total (like it does with NJS set to false) - bearing in mind I haven't looked at your code changes yet so haven't seen how you are doing calculations so might be missing the point here.   But as I said I would never use it with NJS set to true and all the original statistic / results methods were built with this being set to false as that was how they worked, so just ignore me!!

RFBomb:
Original methods relied on `NJS = false`. If job was cancelled or NJS = TRUE, no results would generate. 
This essentially parses the log as it comes in to generate results. 

But the results can be inaccurate, since NJS reports true values, while this is assumed based on the log. 
Without the `/V` verbose option enabled, skipped files and folders are simply not shown at all, so I'm assuming skipping files if it was a copy operation. But thats all that can really be done.

I've done more testing with `/V /NJS` and think this last commit reports fairly accurately, but turning on /V and NJS seems really dumb to me, don't know why anyone would do it. In any case, the estimate will be more accurate based on a Verbose log. 

I don't want to go through the effort of adding the estimated results to the  log lines, if someone else wanted to chip in to do so they can by all means issue a PR for it once this is merged.

PCAssistSoftware:
> but turning on /V and NJS seems really dumb to me, don't know why anyone would do it. 

I completely agree, as you say no reason at all to use either of those switches

Happy to merge unless you need to do anything else?

RFBomb:
`/V` can actually be quite useful, as it shows more detailed file information and which files already exist in the destination. 
But NJS is kinda a dumb one in my opinion.

Note: I also updated how the 'NewDir' ProcessedFileInfo object is created, using regex instead of the parsing, since I found that all directories (including *EXTRA) were being created with the 'NewDir' string in the object.  I kinda wish the 'OnFileProcessed' was split into OnFileProcessed and 'OnDirectoryProcessed' for that, but since it wasn't not much we can do about it without breaking things for previous consumers ( It could be done with a new event specifically for `OnDirectoryProcessed`, but the OnFileProcessed would need to fire anyway)

Yes, I'm satisfied with it.

RFBomb:
Oh, also this obviously doesn't support estimates for languages other than english. 
I can mock that up quickly for future impelementation in the Configuration prior to merge if you want

PCAssistSoftware:
Never really used /V - I will try it now out of curiosity if nothing else

I use /NJH to get rid of unwanted header information in my GUI, and then parse the results to remove bits from summary I don't want - works well for me - as you say would never advise using /NJS

RFBomb:
the Configuration is a simple class, let me set it up for this  before merge

PCAssistSoftware:
> Oh, also this obviously doesn't support estimates for languages other than english. I can mock that up quickly for future impelementation in the Configuration prior to merge if you want

Up to you? if you want to do it before I merge it then fine with me, but if not happy to merge now without it?

PCAssistSoftware:
I see what you mean about /V - I need to make more use of that myself I think as handy to see which files already exist in more detail - thanks

RFBomb:
Ok, I updated RoboSharpConfiguration with a few properties to identify the various log lines. 
Additional language support can now be added to the parser by  updating the `defaultConfigurations` dictionary with the appropriate values.

Too bad we can't get this information directly from robocopy itself or a config file RC uses. If we could, that'd be the ideal way to do it. But I don't know that we can. Either way, it all appears functional and ready for merge, please do some more basic testing first though.

`LogParsing_MismatchFile `and `LogParsing_FailedFile` both aren't verified, since I don't actually know how to force those conditions to occur in order to verify the strings.

PCAssistSoftware:
Excellent, looks good.

Just ran several tests here and all seems fine, so I will merge it shortly.

Usually language specific information for Windows binaries is stored in the associated .MUI file e.g. Robocopy.exe.mui which is located in directory relevant to language e.g. C:\Windows\System32\en-US.  The language strings are often stored in registry as well.  Some time back I had a tool which would view these files as I had to pull out information for another project I was working on in different languages, but that was a while back so will have to look back through my notes and see if that will help us.  That's a job for another day :)

PR #130, opened 2021-12-31
Introduce ProgressUpdateEventArgs

RFBomb:
Expose the result estimates that are being tracked by the ResultsBuilder via a new event - Allows the library to expose the current estimates instead of being performed by the consumer

I think this would be a good addition now that we are doing the math in the resultsbuilder anyway. 
But, I'm torn on how to implement this. 

Currently, this is set up to produce a new RoboCopyResults object and wrap it in an event. Which will result in a lot of ResultObjects floating around.

The other idea was creating an object that implements NotifyPropertyChanged, which is raised whenever the event fires. Much less objects, but may not adhere to traditional event based logic. 

All depends on how its going to be consumed, or if we want to expose the counts during the progress. (which is why this is a new PR). If this is accepted, RoboQueue would have to pass this along as well, but I would just implement its own ResultsBuilder to provide this estimate instead having each RoboComand being subscribed to.


PCAssistSoftware:
The idea and logic certainly makes sense to me, not sure how much input I can give as you are the expert here, what you are doing is impressive and beyond my C# skills i'm afraid, but happy to help as much as I can with testing and discussing ideas etc.

RFBomb:
Yea this is definitely more an idea than anything at the moment. I wanted
to provide it to, for example, be able to provide a way to update progress
bars automatically during RoboQueue mostly. RoboCommand or RoboQueue could
have a new property which is basically a results object that shows the
current estimates, rather than consumer doing the math based on
OnFileProgress events.

In hind sight, all the properties in the results builder that were added in
the last merge can be condensed  into Statistics objects as well, since
that’s all they are building anyway.

PCAssistSoftware:
> Yea this is definitely more an idea than anything at the moment. I wanted to provide it to, for example, be able to provide a way to update progress bars automatically during RoboQueue mostly. RoboCommand or RoboQueue could have a new property which is basically a results object that shows the current estimates, rather than consumer doing the math based on OnFileProgress events.

Now that idea I really like as will make things so much easier!

RFBomb:
I think this is much cleaner. 
I moved all the logic from #129 into a new object, added events to the Statistics object, then created a new event in RoboCommand to declare that an estimator is publicly available to be listened to.

Results Builder Updated -> ProgressEstimator updates its Statistics objects -> Those Statistics Objects raise their events

RoboCommand is pointing to ResultsBuilder's ProgressEstimator, meaning the live results can be actively listened to.

PCAssistSoftware:
Looking really good - will make showing progress so much easier than it is now

PCAssistSoftware:
I am liking that very much, just done a couple of tests and works really well

RFBomb:
When you have a chance, check it out
I've already tested it by putting it onto the Progress tab. Works great. 

Statistics now has other events that can fire, and a new EventArg that shows the new,old, and difference from the values. 
In prep for ProgressEstimator use with RoboQueue, I have provided that event arg a way to add the difference from old to new value to the statistic, and allow binding a ProgressEstimator to another ProgressEstimator to simplify everything in RoboQueue.

RoboCommand now has a new event to declare when the ProgressEstimator is available for binding. This occurs once the Start() method actually starts the process (Source/Destination checks have succeeded and ResultsBuilder is created), then points to the ResultsBuilder's ProgressEstimator.

RFBomb:
I also put a slew of ToString() methods into Statistic for consumers to decide how they want to display their data, if for example they only care about a single value. I'm using the new ToString() overload for displaying to the textboxes with new lines instead of commas.

PCAssistSoftware:
Think we just cross-posted ;)

I gave it a try as soon as I got push notification from GitHub and it worked perfectly for me too, fantastic work, a really useful addition to the library which will make it so much easier than having to do all the various calculations I was doing in OnFileProcessed and OnCopyProgressChanged

Let me know when you want me to merge this for you

RFBomb:
FYI, the bytes is only updated during OnFileProcessed OR when a copy has succeeded. The idea being switching files should determine how last one was processed if it wasn't already processed by a 100% OnCopyProgressChanged event.

PCAssistSoftware:
> FYI, the bytes is only updated during OnFileProcessed OR when a copy has succeeded. The idea being switching files should determine how last one was processed if it wasn't already processed by a 100% OnCopyProgressChanged event.

That make sense

RFBomb:
All Set

MarkECrum:
Nice.... However, I am having trouble implementing it... I tried various way to get it to spit out the values in the event.... but no luck.... Can you guys update the main page with usage examples please... pretty please?
I don't have access to a "Dispatcher" frankly I have search and don't even know what it really does other than thread safe communication on the main UI.... I use a class written to do this. So the example is on the new event. It prints out something like this... but never updates the values: 
   "Total: 0, Copied: 0, Skipped: 0, Mismatch: 0, Failed: 0, Extras: 0" All info that I would like to use of course!

private void RoboCopy_OnProgressEstimatorCreated(RoboCommand sender, RoboSharp.Results.ProgressEstimatorCreatedEventArgs e)
        {
            txtDirStats.SafeInvoke(d => d.Text = e.ResultsEstimate.DirStats.ToString());
            txtFileStats.SafeInvoke(d => d.Text = e.ResultsEstimate.FileStats.ToString());
            txtByteStats.SafeInvoke(d => d.Text = e.ResultsEstimate.ByteStats.ToString());
        }

PCAssistSoftware:
Have you taken a look in the sample BackupApp as it is demonstrated there?

```
 private void Copy_OnProgressEstimatorCreated(object sender, Results.ProgressEstimatorCreatedEventArgs e)
        {
            e.ResultsEstimate.ByteStats.PropertyChanged += ByteStats_PropertyChanged;
            e.ResultsEstimate.DirStats.PropertyChanged += DirStats_PropertyChanged;
            e.ResultsEstimate.FileStats.PropertyChanged += FileStats_PropertyChanged;
        }
```

MarkECrum:
I did implement the event hooks but did not look at it.... Doing it now, thanks!

RFBomb:
Yea, a usage example would be required for this one, I agree. It’s not
exactly intuitive, since it’s not exposing the object until the run has
started, since that is when the object is created.

Basically you want to set up an event like you did to catch that it has
become available, or check if the RoboCommand estimator is not null after
starting the command.

Then, you subscribe to the event of the statistics object that are provided
by the estimator object.

—-

private void RoboCopy_OnProgressEstimatorCreated(RoboCommand sender,
RoboSharp.Results.ProgressEstimatorCreatedEventArgs e)
{
e.ResultsEstimate.FileStats.OnPropertyChanged += (o,e2) =>
txtFileStats.SafeInvoke(d => d.Text = o.ToString());
}

—-

The dispatcher there was just how I discovered WPF wanted it done to invoke
it against the UI thread.

MarkECrum:
Thanks to both of you! PCAssistant got my going, I went to the sample app (duh)... There is needed some formating I think... look at my picture below...

Sample

RFBomb:
It looks like you used the ‘ToString(bool,bool,string,bool) method, not the ‘ToString’ method. 

ToString() returns the same thing the old version did, while the one with the additional parameters allows customization (to for example show/hide the identifier for total/skipped/etc). It requires some delimiter be specified, if none is specified it will turn out as you show. Specify a decimeter and it should solve your issue.

RFBomb:
I’ll put through a PR that updates this some usage description and a default delimiter “, “

PR #128, opened 2021-12-23
RoboQueue Object - Work with multiple RoboCommands

RFBomb:
Now that #127 is merged, work can begin on an object that facilitates running multiple commands in succession, rather than every consumer of the library building their own implementation.

This PR will need to be heavily tested prior to merging, and should have input from as many folks as possible regarding how they use the library in such a scenario.

Relevant thread:
#126

Issue #120, opened 2021-12-09
IsRunning = False, even if the copying process has not yet been completed


JoE-Seeve:
What is the meaning of the different flags for IsRunning, IsPaused and IsCancelled?

I have observed the following:
- IsRunning == True -> If the copy process is running
- IsRunning == True and IsPaused == True -> If the copy process isn't stopped and the process is paused
- IsRunning == False and IsCancelled == True -> If the copy process is running but a cancelation was requested
- IsRunning == False and IsCancelled == True -> If the process has completed but was cancelled

I would actually expect the following:
- IsRunning == True and IsCancelled == True -> If the copy process is running but a cancelation was requested
- IsRunning == False and IsCancelled == False -> If the process has completed but was cancelled

I only noticed this when I copied to a very slow USB stick and the message was constantly displayed that the copying process was no longer running, but after that there was no "OnCommandCompleted" signal for minutes. Could someone please explain to me how you can interpret the signals to clearly show the state of the copying process? Thanks in advance.

JoE-Seeve:
After more time to test the RoboSharp signals, I believe that there is more of a bug here. I found the following problems:
- IsCancelled is set to TRUE when the copy function is started for the first time if the copy process is stopped. But there is no code in RoboCommand.cs that sets IsCancelled back to FALSE.
- At the moment there is no signal that indicates that the copy process is running or is being ended after a STOP command. IsRunning is set to FALSE if IsCancelled = TRUE. The copying process can still run. Since IsCancelled is never set to FALSE, this flag does not indicate when the copying process is ended.

It should work with the following adjustments (at least for me):
- IsCancelled is set to FALSE every time a copy process is started, indicating that the current or the last copy process has been canceled. Most users will not notice this behavior because they reinitialize the class for each copy process, just as it is described in the example (RoboCommand backup = new RoboCommand();).
- A new CancellationRequested signal indicates that a cancellation or stop has been requested. This flag is set to TRUE on STOP and back to FALSE on completion of the copy process (IsCancelled does not need to be changed so that existing apps should not have any problems).

PCAssistSoftware:
Sorry for delay in responding, when I get a moment in the coming days I will look into what you are saying and comment further.  I use those flags in my apps and have never noticed them returning wrong values though, but will certainly investigate

JoE-Seeve:
Thank you very much for your info. I would like to present my solution via a pull request, but I still need time to read the help and then be able to use it.

[...]

PR #133, opened 2022-01-05
Robo command updates

### LINKED ISSUE #102: IsRunning = True even if errors occur

RFBomb:
Bring in various fixes from #128 for RoboCommand object

- This should also resolve #102  -- Needs Verification 

**Update Dispose and Stop methods**
- Dispose and Stop previously looked at the HasExited private bool.
Since that is private to the object, and not describing the process itself, these have now been updated to evaluate Process.HasExited instead.
- Stop() is now updated so that it always reliably exits.

**Disable NJS Default**
-  This was turned on for testing and we forgot to turn it back off

**Update Start() Method**

- Reset `isCancelled`, `isPaused` bools when the command is started.
- Always run the continuation task to allow the OnCommandCompleted event to fire and to allow the Stop() method to dispose of the process.
- Move the Stop() command in the continuation task prior to raising the OnCommandCompleted event to ensure that all cleanup within RoboCopy has completed first.

### DIFF
--- a/RoboSharp/RoboCommand.cs
+++ b/RoboSharp/RoboCommand.cs
@@ -204,7 +204,6 @@ namespace RoboSharp
                     {
                         OnError(this, new ErrorEventArgs(data, parsedValue));
                     }
-
                 }
                 else if (!data.StartsWith("----------")) // System Message
                 {
@@ -223,8 +222,6 @@ namespace RoboSharp
                 }
             }
         }
-        
-    
 
         /// <summary>Pause execution of the RoboCopy process when <see cref="IsPaused"/> == false</summary>
         public void Pause()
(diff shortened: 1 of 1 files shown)


PCAssistSoftware:
Do you need me to test anything here, or are you happy for me to merge?

RFBomb:
This is fine to merge, but Joe over in #120 should check to see if the problem still exists after the merge since he knew what he was looking at.

This was primarily reorganizing the constructors, adding the Name property, and ensuring Stop() always runs. This has been tested within #128 already

PCAssistSoftware:
@RFBomb - see response in #120 from Joe re this fix

PR #135, opened 2022-01-05
Introduce ReadOnly Interfaces

RFBomb:
When I added in the StatType enum, I forgot to specify a StatType for the RoboCopyResultsList stat objects. 
This fixes that by introducing a new parameter to the static methods. 

Name property will also now raise NotifyPropertyChanged event, and `public void AddStatistic(PropertyChangedEventArgs eventArgs)` have been adjusted accordingly. 

Two issues / concerns remain:

 - `public void AddStatistic(Statistic stats)` allows adding any two statistics objects together. Obviously this means that you can add a FILE stat and a DIR stat. This library won't be doing that, but should this method only allow adding of like types?
     - My idea is that stat with type `unknown ` can add any stat to it, but File/Byte/Dir types will ignore dissimilar types.
     - OR, allow any stat object to add to another Stat object, but if types are dissimilar update the type to `unkown` / throw error. 

Same concern exists for public Subtract method.

### DIFF
--- a/RoboSharp/Results/ProgressEstimator.cs
+++ b/RoboSharp/Results/ProgressEstimator.cs
@@ -7,21 +7,6 @@ using System.Text;
 namespace RoboSharp.Results
 {
     /// <summary>
-    /// Provde Read-Only Access to the a <see cref="ProgressEstimator"/> object
-    /// </summary>
-    public interface IProgressEstimator
-    {
-        /// <inheritdoc cref="ProgressEstimator.DirStats"/>
-        IStatistic DirStats { get; }
-
-        /// <inheritdoc cref="ProgressEstimator.FileStats"/>
-        IStatistic FileStats { get; }
-
-        /// <inheritdoc cref="ProgressEstimator.ByteStats"/>
-        IStatistic ByteStats { get; }
-    }
-    
-    /// <summary>
     /// Object that provides <see cref="Statistic"/> objects whose events can be bound to report estimated RoboCommand progress periodically.
     /// </summary>
     /// <remarks>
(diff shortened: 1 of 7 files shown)


PCAssistSoftware:
Personally think you should only allow adding of the same type.  Not sure why anyone would want to add different anyway, but probably best to not allow it.  So your first bullet point would be my vote.

Question though - what are "unknown" stats, surely we only ever have file / dir or byte?

RFBomb:
Yea, I added unknown for the parameterless constructor.
Removing the parameterless constructor would potentially break applications that go to update to the latest DLL, or at the least cause them to use the constructor that specifies the type. 
I think thats fine personally, since the chance of occurring is unlikely, but that was my line of thinking.

PCAssistSoftware:
Okay, understood now

RFBomb:
So, yay or nay to keeping 'unknown' ? (And hiding parameterless constructor)

PCAssistSoftware:
My thought would be get rid of it, especially if we only allowing "adding" of same types anyway, I think "unknown" just confuses things, it certainly did me

PCAssistSoftware:
Let me know when you want this one merged

RFBomb:
Ok, So I believe this is mergable. 
In the strive for thread safety (and the realization that ResultsList and Estimator values were able to be modifed publicly outside the control of the class), I made interfaces and provide them for the statsn objects. 

I did not update the RoboCopyResults objects to use the interface, since that has been long-standing and anyone thats used this library would need to revise it to accomodate the interface. So to avoid that huge headache, that class is untouched. But all the new stuff is now using the interface to present a way to read and not modify the results by default.

Issue #136, opened 2022-01-12
Net40 Support


RFBomb:
Formal request to drop Net40 from target frameworks. 

Net40 has been deprecated and has not received security updates since 2016. It also prevents use of Async methods, along with await.

it would also remove the need for the `#if net40` statements, since nearly all of the other targets (save for edge cases easily worked around) have support for nearly all the same methods available. 

My reason for this request is RoboQueue. During my testing, what I’ve been seeing is that occasionally RoboCommands get stuck in a ‘waiting to run’ state while the RoboQueue method is in its sleep loop, meaning it is at a standstill that will Never complete. Using `await Task.delay` Frees the thread up to do work, such as starting the robocopy start command. 

What I propose:
- take all the improvements from RoboQueue except the BackupApp changes and RoboQueue itself and submit them as PRs. This bring any changes I did to optimize /improve/ clarify them to all items. 
- Create a new fork that can continue Net40 support bug fixes if needed. 
- One last Nuget Package that includes Net40 support, put it into the readme. 
- Then remove it from dev

PCAssistSoftware:
All of that makes perfect sense to me, and I fully agree, But let's see what @tjscience has to say on the matter first :)

RFBomb:
That’s why I made it a new issue! 
(and just in case any Net40 projects around still exist that are using this have a reference thread to look at if they pull an incompatible release)

RFBomb:
Net40 branch created.

Issue #105, opened 2021-11-03
Cannot stop PriorityScheduler threads


wanga7:
The threads launched in PriorityScheduler run forever until my program stops. It would be nice to provide a mechanism to manually stop those threads, or starts them on-demand. Thanks!

RFBomb:
RoboCommand.Stop() should work (as long as you still have the object reference), but Stop() currently has a check for the RunHours property. 

I've wrapped the Stop() command in my helper class so that I set RunHours = null before calling Stop(). This stops the robocopy thread immediately, if it is running. 

Also, my PR #107 has introduced a new property `StopIfDisposing` that should do what you want when set to TRUE. If the RoboCommand object is disposed, and the RoboCopy process is still running, it kills it. 

That said, I'm not using PriorityScheduler right now, so this would have to be verified.

wanga7:
I did some testing locally and here're my conclusions:

1. The problem I mentioned above is not about "roboCommand process" not stopping, but "the task scheduling threads allocated by PriorityScheduler" not stopping until program exits.
2. If you take a look at the PriorityScheduler class you'd see it creates the threads but there's no way to end them.
3. Swapping out the PriorityScheduler with the .Net default one (TaskScheduler.Current) seem to solve the problem.

PCAssistSoftware:
@wanga7 - can you please test again after all the changes made from RFBomb in #107 have been merged and taking into account comments in reply above to see if still an issue?  thank you

wanga7:
See my comment above: it's unrelated to RFBomb's problem and his fix won't solve my problem here.

PCAssistSoftware:
Okay thanks for confirming, feel free to submit a PR if you have a fix then

RFBomb:
Are these threads sitting there consuming resources? Or do you wish to just have them killed off if all RoboCommands have finished executing? Best I can tell, those threads should just be sitting idle if no tasks are scheduled.

RFBomb:
@tjscience , @PCAssistSoftware , @wanga7 I just ran into something peculiar when implementing RoboQueue into my application. It appeared as if RoboQueue was not working properly in my scenario, and I think the PriorityScheduler may be at fault.

My Application queues up 18 RoboCommands, stuffs them into a RoboQueue, then starts them all. First in ListOnly mode, then once they all complete, in copy mode. For List-Only mode, I start all commands at same time, but for copy mode only 1 runs at a time. This is all fine and handled by RoboQueue.

[...]

Issue #103, opened 2021-10-31
Unicode characters from Process


JoE-Seeve:
Special characters of a language (for me the special characters of the German language) are not returned by RoboSharp in the different events.

The following could help to solve this (RoboCommand.cs - Start):

 process.StartInfo.StandardOutputEncoding = System.Text.Encoding.GetEncoding(CultureInfo.CurrentCulture.TextInfo.OEMCodePage);

RFBomb:
Should be resolved

Issue #83, opened 2020-02-20
Robosharp can't work with non-english chars


Dan-Baruch:
When trying to set source / destination of copy that contains a directory with Hebrew characters the callback events does not return proper file name and that can't be worked with.

giesalabs:
it don't copy the files right?

could be a charset problem in the arguments?? there are many example on google: https://stackoverflow.com/questions/2627891/does-process-startinfo-arguments-support-a-utf-8-string

in this link they suggest to use Arguments = Encoding.Default.GetString(Encoding.UTF8.GetBytes(args)); 

so in RoboCommand.cs at line 288 maybe is possible to try this line
` process.StartInfo.Arguments = Encoding.Default.GetString(Encoding.UTF8.GetBytes(GenerateParameters()));`

but i never try a different charset in the Arguments and i'm not a Robosharp developer, so the developers team should try out 
anyway just for curiosity, could you try the code line?

Dan-Baruch:
I've tried this line long long ago to no avail. I ended up not using this library though, as I couldn't find a way around that

giesalabs:
could be a robocopy problem?
can you call the robocopy by prompt and backup files?

Dan-Baruch:
Well, I ended up not using roboCopy. I used to be able to make it work only if I were to write all the paths to a log file, but that also had it's limitations.

giesalabs:
in my test i can see that the problem is not in the line 288 (RoboCommand.cs) because this line will not be executed when there are special chars in the destination path
the code find some error much time before, now i think that there is somewhere a RoboSharp bug

giesalabs:
after more test it works with non english chars, in the test i did the mistake to use chars not allowed in a windows path

Dan-Baruch:
Well, I know that if my target ends with a single directory in Hebrew characters it was ok. If, how ever, I had something like:
C:\EnglishPath\EnglishPath\HebrewPath\EnglishPath\HebrewFile
It failed.
Of course, there's no relevant to how many directories there are, it's only a matter that there's a file in Hebrew inside a directory in Hebrew.

JoE-Seeve:
I used the following line to initialize the CodePage to match the operating system. That worked for me (e.g. with German special characters). This should make it possible to get all characters of the currently installed language back from RoboSharp.

                process.StartInfo.StandardOutputEncoding = System.Text.Encoding.GetEncoding(CultureInfo.CurrentCulture.TextInfo.OEMCodePage);

[...]

PR #137, opened 2022-01-12
Introduce Interfaces, Restructure Files, Some Optimizations

RFBomb:
This is PR is comprised of the numerous changes made while developing #128, but without RoboQueue specific items. 
My goal with this PR is to bring in all the bug fixes, tweaks, improvements, and structure changes prior to creating the PR for #128.

This PR is also what I feel should be the last commit for Net40 support, as laid out in #136.

Essentially what I did was copy my RoboQueue folder in its entirety, create a new branch based off the current Dev branch, then overwrite all files with the files from RoboQueue. I then went through each file and created commits to showcase how each file changed, so that the individual changes can be viewed and compared. 

What this encompasses:
- Moved all EventArgs objects out of the main folder / parent classes and into a new folder for Event Args. I named it EventArgObjects to avoid autogenerated files being `RoboSharp.EventArgs` namespace, which wipes out the ability to call up the `EventArgs` class.
- Moved all Interfaces I created and commited previously out of their parent class's files and into the Interface folder where they should have been originally. (Woops)
- Some EventArgs have been introduced.
- BugFixes and Optimizations have been made on various files, the largest of which was the Statistics file.

I tried to contain the changes for each file within just a few commits that can be reviewed (and described) individually to allow them to showcase each change.


PCAssistSoftware:
Excellent work , let me know if you want me to merge this now or later?

RFBomb:
Hold off on the merge for now, I want to verify that @tjscience  is fine with the moving the EventArgs and give him a chance to review.

PCAssistSoftware:
Okay, that's fine

tjscience:
Hi guys. Everything looks really good here. Thanks again @RFBomb for all of your work here. I am also in favor of splitting off a last release and nuget package for .NET4.0. We actually did the same thing for .NET3.5 awhile back, so this will be consistent.

RFBomb:
OK, I just went poking through the issue lists seeing if there was anything else worth messing with, including #105 (which I don't see why its an issue if threads are idle, but can understand the desire for the request, just don't know enough about it to even think about such a change). 

I put in the recomendation from #103 that was suggested also as a fix for #83. It is in the configuration class now, so by default it will use system language, but can be set null if a consumer needs it. (Previous functionality it was never set, so it remained null)

This should introduce the multi-language support there, and with the other changes made to this class, expanding language support should be doable should someone need to.

I;ve done some test copies, and since this fix is also what Joe did in #103  for german language support, I believe it should be fine.

As far as I am concerned, this is mergable. I've only left RoboQueue specific items out of this PR.

PCAssistSoftware:
Okay, happy to merge it for you now if ready?

RFBomb:
Yes, this is ready for merge. I'll start prepping RoboQueue actual PR once I merge this into that

PR #138, opened 2022-01-13
DropNet40, Add RoboQueue

RFBomb:
This should be the last of the changes from #128 

This Drops Net40 support ( so prior to merging in, Dev should be forked to have a Net40 branch ).
Then it adds RoboQueue and all associated changes.


PCAssistSoftware:
Created NET4.0 branch - please ensure this looks okay as never done that before - if it is okay I will then do this merge for you

RFBomb:
Should be fine, I believe you basically needed to just create a new branch and remove tracking from dev

PCAssistSoftware:
Yes I just read help section and did what it said - will merge now

RFBomb:
I don't know if your Readme changes over in #128  made it into this FYI

PCAssistSoftware:
> I don't know if your Readme changes over in #128 made it into this FYI

Don't think I had made any readme changes in that one? all changes I made I can see in dev so should be fine.  Obviously we need to document all the new changes at some point so it will need a lot of work doing anyway.

Issue #121, opened 2021-12-10
ExcludeFiles and ExcludeDirectories - possible improvement?


PCAssistSoftware:
For both of the above you can pass multiple values but if a file name or path contains spaces then they need to be encased in quote

e.g.

`c:\files c:\pictures`

or

`c:\files "c:\my photos"`

etc

If you pass a value which has spaces but don't enclose it in quotes then it doesn't work - as you would expect.

Can we do anything to parse the value you are passing to either of these to ensure it is quote as necessary?

Not a simple task and may not even be possible as would need to be able to work out what is a space between two separate entries and what is a space in the middle of a path e.g. 

`file1.txt file2.txt `

would be obvious as space is between two separate file names, but 

`file.txt my file.txt`

wouldn't be obvious as one space between file names and one space 'in' the second file name

Thoughts / ideas? or is this just going to have to rely on advising user how to format values rather than being able to enforce or help with it?

RFBomb:
The only foolproof way to implement this within the library would be to introduce a List<string> for exclude files / directories, then loop through the list when building the command. That way, each string in the list is processed individually, and is sensible for consumers to implement. Then just use a 'LoggingOptions.WrapPath' method to process them

PCAssistSoftware:
Thanks for reply - yes I thought that may be the answer, at least like that you can ensure all items in list are quoted correctly if contain spaces etc

Probably something best handled in the calling application rather than directly in the library then, as I do for things such as file filters

RFBomb:
Rather than updating the current methodology of just storing the list in a string format, the only thing I can think of at the moment is introduce a new method that accepts IEnumerable<string> as  the parameter, loops through as I described above, added each item to the desired list of inclusion/exclusions. So we would need a method for each property to be able to do that. THis way, it can accept a list, but will do the conversion automatically, ensuring we don't break existing applications or cause ourselves unneccessary headache by converting it to a list then being backwards compatible.

PCAssistSoftware:
Hi @RFBomb 

Just noticed that in latest version you have added in `SelectionOptions.ExcludedFiles.Add` and `SelectionOptions.ExcludedDirectories.Add`

Not sure at what stage of implementation you are with these but just been giving them a try - apologies if testing before ready!

Currently they have same issue I had when originally opening this issue e.g. the handling of a space in a file or path name

e.g.

Adding 

```
job2.SelectionOptions.ExcludedFiles.Add("myfile.txt")
[...]

PR #139, opened 2022-01-13
Job Options

RFBomb:
Since JobOptions has support requested, this PR is meant to resolve that request. 

Heavy WIP


RFBomb:
To Do:
Add internal override to SAVE that allows saving the job file, since only saving to a job file should include the JobOptions parsed string, which needs to be at the end of the command to work properly.

RFBomb:
@PCAssistSoftware  - Regarding #121 - If this is going to be implemented, now is the time to do it. 
See last commit.

PCAssistSoftware:
> @PCAssistSoftware - Regarding #121 - If this is going to be implemented, now is the time to do it. See last commit.

Okay, makes sense, happy to test anything you come up with, but not sure how to implement?

RFBomb:
We need a button on the form that opens up a FilePicker window. It needs to only show ‘RCJ’ files. 

We also need a text box to show that filepath, so it can be saved to /read from.

JobFile can constructed by a few different means, basically anything that can produce a IEnumberable<string>  And implements IRoboCommand.

LoadCommand can be used to display the parsed JobFile’s options via IRoboCommand

the Save method isn’t built yet, but we will need a button for it. 

As far as integrating it into RoboCommand, I’m planning to create a ‘MergeJobFile’ method to combine these job file options with an existing command’s options. Will be easier than having a list of jkbfiles to evaluate, which would require new  batch of CopyOptions, SelectionOptions, etc that override the properties to also do a List<JobFile>.Any( X ) check. It just gets over complicated. So I don’t want to support adding and removing JobFiles, only Merging their options (which is what RC does once run)

RFBomb:
As far as #121, I put these in already,  but it’s not tested. the existing an BackupApp can test this since the old properties are redirected to the new list replacements. 

but I’m wondering if it should be marked as ‘Obsolete’ since it’s still functional, just redirected. (I think it makes sense, just it could be hidden by default)

PCAssistSoftware:
How does this work for multi-job? as the standard RoboCopy job file just contains a single job I believe? so if you wanted to save jobs from a multi-job you would have to save one for each job? and then load one for each job?  

I appreciate that is how RoboCopy does it but surely for your excellent multi-job / roboqueue implementation it would be better to have our own methods which can save all of the current job queue information to one file (perhaps in XML format for easy parsing?)

RFBomb:
RC allows saving a command to a job file.it also allows selecting multiple job files and combing their options into a single run. 

 This is meant to build in that functionality. So for RoboQueue, you could `RoboQueue.Add((IRoboCommand)JobFile);`,
Or do a ForEach at RoboQueue and merge in a job file to all jobs in the queue. 

As far as your suggestion of multiple jobs, I don’t really want to write that, as it can get obnoxious quick, and it’s also proprietary to this tool. 

What we could do is create a method that points to a DirectoryInfo object, and creates a JobFile for every RCJ file in the folder, that would be trivial.

I plan to create some job files and embed them into my project and return them via StreamReader using reflection, for example, then pass those into a RoboQueue object. 

To me, writing a proprietary method and storage to save multiple jobs isn’t worth it when this functionality can be easily Automated to do the same thing

PCAssistSoftware:
Okay

PCAssistSoftware:
Okay - so if I had 20 jobs in my job list then it would save 20 RCJ files each containing 1 job? rather than 1 file containing all 20 jobs?

Seems messy to me, so think we will just to agree to disagree on this area.

RFBomb:
Yea, it would save 20 files, as each RoboCommand saves out to its own RCJ
file when using RoboCopy Jobs built in functionality.

If that’s a problem, you feel free to write  something that does it how you
want.

The way I would do something like that is write all the files to a
temporary location, the write a new one to the requested destination,
simply looping through all files generated and appending lines to the
destination file, with some delimiter.

If using XML, write methods to serialize/deserialize RoboCommand, which is
likely the easier method.

RFBomb:
You can run XSD Tool against the classes you want to serialize, and I wrote An extension for VS to assist with doing the serialization. 

Some extensions that may help

https://marketplace.visualstudio.com/items?itemName=PascalCabanelxsd2codecom.xsd2codeCodegeneratorfromXMLschemas

https://marketplace.visualstudio.com/items?itemName=RBrenckman.XsdCustomTool275 
( I made that second tool to include the serializer and deserializer methods, but post processing of the file it generates is still required.)

feel free to integrate it, but I’ve had enough headaches with XML in my other project (which is why I wound up designing the extension) to avoid it if I can.

PCAssistSoftware:
Thanks, will have a read later

PCAssistSoftware:
My apologies if I came across negative yesterday, just giving my opinion, no offence meant.

Still here to test anything you implement, and will look more when I have time at your useful links and think how I want to achieve my requirements of storing ALL jobs in ONE file complete with job name etc

:)

RFBomb:
No worry. 
I’ll open up a new PR draft for that with all the xml stuff generated from my extension. 

It’ll add XSD schema files, and potentially require heavy tuning to get it to work. Plus attributes galore. You’ll see what I mean once I open it.

RFBomb:
OK, I tried my hand at it this morning, big changes would be required to make the XML thing happen. 

see output from XSD tool:
```
C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools>XSD "C:\Users\robbe\source\repos\RFBomb\RoboSharp\RoboSharp\bin\Debug\net45\RoboSharp.dll" /c /o:"C:\Users\robbe\source\repos\RFBomb\RoboSharp\RoboSharp\XML_Files\"
Microsoft (R) Xml Schemas/DataTypes support utility
[Microsoft (R) .NET Framework, Version 4.8.3928.0]
Copyright (C) Microsoft Corporation. All rights reserved.
Error: There was an error processing 'C:\Users\robbe\source\repos\RFBomb\RoboSharp\RoboSharp\bin\Debug\net45\RoboSharp.dll'.
  - There was an error reflecting type 'RoboSharp.Debugger'.
  - Cannot serialize member 'RoboSharp.Debugger.DebugMessageEvent' of type 'System.EventHandler`1[[RoboSharp.Debugger+DebugMessageArgs, RoboSharp, Version=1.2.5.0, Culture=neutral, PublicKeyToken=null]]', see inner exception for more details.
  - System.EventHandler`1[RoboSharp.Debugger.DebugMessageArgs] cannot be serialized because it does not have a parameterless constructor.

If you would like more help, please type "xsd /?".

C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools>XSD "C:\Users\robbe\source\repos\RFBomb\RoboSharp\RoboSharp\bin\Debug\net45\RoboSharp.dll" /c /o:"C:\Users\robbe\source\repos\RFBomb\RoboSharp\RoboSharp\XML_Files\"
```

the tool needs to run directly against an XSD file, or a DLL assembly. when against a DLL, it tries to generate files for all assemblies within the dll (meaning each public class). CopyOptions errors out due to the IEnumerable<string> FileFilter, and any other class that is public may have issues that will need to be addressed before it can serialize the class. Thats a lot of work that I'd rather not put in place, though I can understand why you would want that functionality.

You're best bet is to manually create an XSD file for all the classes, then run my extension against them to generate the serializeable class and the helper code to serialize/deserialize it to/from disk.

Unless the other extension I linked can assist here, I have no Idea as I've never used that extension.

PCAssistSoftware:
Thanks for looking, I appreciate it. 

Not to worry, I will look into it myself when I have more time.  I have done previous apps using XML to store data so will look back to see how I did them because I certainly didnt have to go through any of those steps from what I remember, think I just used XmlTextWriter and defined the schema in code, below was just a snippet from an app where I was storing internet favorites in XML, I have a more advanced one somewhere storing lots of information so will dig that up and take a look

```
 Dim tw As XmlTextWriter = New XmlTextWriter(linksName, System.Text.Encoding.UTF8)
        tw.Formatting = Formatting.Indented
        tw.WriteStartElement("Items")
        tw.WriteAttributeString("version", "1.0")
        tw.WriteAttributeString("application", Application.ProductName)

        tw.WriteStartElement("Homepage")
        tw.WriteElementString("Home", HomePage)
        tw.WriteEndElement()

        tw.WriteStartElement("Favorites")
        For i = 0 To lvwLinks.Items.Count - 1
            tw.WriteElementString("Favorite", lvwLinks.Items(i).Tag.ToString(), lvwLinks.Items(i).Text.ToString)
        Next i
        tw.WriteEndElement()
        tw.WriteEndElement()
        tw.Close()
```

But please don't waste any more of your time on my requirement, I do appreciate your advice though :)

RFBomb:
The project that I was working on that prompted me to build that extension was dealing with XML files that are exported from external sources, so I had no control over the contents of the xml, I could only build classes around the xml. So I used XSD tool from Microsoft to create an XSD schema from the xml file, then ran XSD against the schema and it generated a class to use. 

The class it generated, as long as the XSD file was tweaked to get it proper, didn’t require a custom xml writer, I could simply use the Serialize/Deserialize in and out the class object. 

But XSD doesn’t automatically generate all members, so I wrote the extension to add them in. (For example, is a parameter less constructor was not provided by XSD, create one that initializes all properties so the dev doesn’t have to. It also adds a few methods to serialize and deserialize to some file path. Basically adds in the fluff to make it usable that XSD neglects

PCAssistSoftware:
Thanks, I will dig up my other project later and take a look, certainly it was storing a lot of information so will be similar to what I will need for this

:)

RFBomb:
@PCAssistSoftware - I think this is just about ready for testing and merge.

You can now Save and Load JobFiles from the BackupApp. 

Lines 294 currently performs a MERGE function, merging the loaded JobFile in with the currently selected options in the Options Tab. To make it Load the file and only show the options that are in the file itself, change line 294 from 
`LoadCommand((IRoboCommand)oldCmd);`
to 
`LoadCommand((IRoboCommand)JF);`

PCAssistSoftware:
Very nice indeed, great work.

Just tested with various settings and apart from couple of comments below is working well :)

Comments:-

1) Personally I think it should be set to JF on line 294 as oldCmd (merge) makes less sense to me, e.g. even source / destination paths don't get loaded correctly when set to oldCmd?  I can understand merging of job options perhaps, but not things like source, destination or name

2) Needs some sort of check / warning in case an empty or corrupt .RCJ file is loaded, as at the moment it just throws a System.NullReferenceException: 'Object reference not set to an instance of an object.' when trying to set the values

On a separate not just in case you didn't see them - there is an issue with cancelling single jobs as per notes [here](https://github.com/tjscience/RoboSharp/issues/120#issuecomment-1013905748)

RFBomb:
Race Condition Fixed in the last commit. Stop() is still exposed, but now the logic for process disposal is hidden by a private Stop(bool) method that is only called after RoboCommand finishes executing or when RoboCommand is being disposed of.
This way, they can call 'stop' no matter how much they want and it will essentially kill the process without disposing of it before other threads finish their work.

RFBomb:
> Personally I think it should be set to JF on line 294 as oldCmd (merge) makes less sense to me, e.g. even source / destination paths don't get loaded correctly when set to oldCmd? I can understand merging of job options perhaps, but not things like source, destination or name

I agree, but I set it to merge so you can see the functionality and for testing purposes.

>Needs some sort of check / warning in case an empty or corrupt .RCJ file is loaded, as at the moment it just throws a System.NullReferenceException: 'Object reference not set to an instance of an object.' when trying to set the values

Agreed. Incoming.

PCAssistSoftware:
> Race Condition Fixed in the last commit. Stop() is still exposed, but now the logic for process disposal is hidden by a private Stop(bool) method that is only called after RoboCommand finishes executing or when RoboCommand is being disposed of. This way, they can call 'stop' no matter how much they want and it will essentially kill the process without disposing of it before other threads finish their work.

Tested and working fine for me :)

PCAssistSoftware:
> > Personally I think it should be set to JF on line 294 as oldCmd (merge) makes less sense to me, e.g. even source / destination paths don't get loaded correctly when set to oldCmd? I can understand merging of job options perhaps, but not things like source, destination or name
> 
> I agree, but I set it to merge so you can see the functionality and for testing purposes.

Perhaps we could have a checkbox for merge? so if checked then it merges (oldCmd), if unchecked it doesnt (JF)? EDIT: and by default it is unchecked
But I still think (if possible) only options get merged - NOT things like source, destination or job name?  thoughts?

> 
> > Needs some sort of check / warning in case an empty or corrupt .RCJ file is loaded, as at the moment it just throws a System.NullReferenceException: 'Object reference not set to an instance of an object.' when trying to set the values
> 
> Agreed. Incoming.

Okay - thanks

RFBomb:
Ok, so the JobFileBuilder is producing a RoboCommand object.
If it can't, it produces a null object, which is where the error comes in.
Should we default always creating a Robocommand, or should we implement a Factory method to be able to return null in case of inability to construct. 

I'm in favor of factory method to return null if unable to construct.

PCAssistSoftware:
> Ok, so the JobFileBuilder is producing a RoboCommand object. If it can't, it produces a null object, which is where the error comes in. Should we default always creating a Robocommand, or should we implement a Factory method to be able to return null in case of inability to construct.
> 
> I'm in favor of factory method to return null if unable to construct.

Also in favor of factory method to return null

RFBomb:
Hows That?

PCAssistSoftware:
> Hows That?

Perfect :)

Happy to merge if you agree?

RFBomb:
I'm happy with it, it seems to have resolved the last few outstanding issues (that we are aware of) and JobFiles are now fully supported. 
The only thing that may require some more testing I think is Saving/Loading/Running `SelectionOptions.ExcludedDirectories` and `SelectionOptions.ExcludedFiles` 

Writing to the strings causes it to be split using `FileFolderNameRegexSplitter`, which hasn't been tested thoroughly yet.

The goal with that regex is to match as many times as possible within the string itself, then split that into a list<string>

For example:
`"*.txt *.doc *.docx"` should translate to `{ "*.txt", "*.doc", "*.docx" }`

The regex may need adjusting. Best spot to test would be a breakpoint on line 108 in SelectionOptions.

PCAssistSoftware:
Okay, will merge now and will add to my list for this week to test further re ExcludedDirectories and ExcludedFiles

RFBomb:
> But I still think (if possible) only options get merged - NOT things like source, destination or job name? thoughts?

If you examine the 'MERGE' method and the summaries I gave them, they detail what gets merged and what takes priority. 

For example, if RoboCommand has a source and destination, and so does the Job File, the Job file's source and destination are ignored for the merge. 

But if job has source and RoboCommand does not, RoboCommand will adopt the job's source. 

The strings that are merged are the ones that combine, for example the attributes. 

You can't specify an attribute identifier in the string twice, so it merges the ones that don't exist in the RoboCommand's String.

PCAssistSoftware:
okay, understood

RFBomb:
Sorry, was just reviewing the thread and didn't think I answered that ><

PR #140, opened 2022-01-16
Fix RoboCommand Start producing Task<Task>

RFBomb:
Also fixed PreCompile Directives not targeting targets_OR_GREATER

Resolve issue from this morning's post in #120 

I was running into the issue of starting too many commands with RoboQueue, which was hanging the TaskScheduler implementation since the tasks were running on threads that were being slept, meaning no more tasks could run on those threads. 

SO I had made the sleep and `Process.WaitForExit` asynchonrous.

Well, I found out that `Task.Factory.StartNew` doesn't natively handle Async Tasks, but `Task.Factory.Run` does.
`Task.Factory.Run` silently detects if you are producing `Task<Task>` and unwraps it automatically. `StartNew `is lower level, and you have to unwrap it yourself. I was unaware of this nuance when I made that change. 
After researching the issue, I found a MS article that explains the nuance of `Task.Factory.StartNew ( async() =>` creates a task that returns an async task, so the continuation task fires as soon as the inner async task has done the first 'await'. This caused RoboCommand.Stop() to be called, and reported results that weren't build yet, crashing everything.

I have added `.Unwrap()` to the end of the StartNew() task, which instead of returning Task<Task> will now simply return the inner task. Tested this morning and it works as it used to, but now doesn't block the entire thread since its waiting asynchronously.

### DIFF
--- a/RoboSharp.BackupApp/JobHistoryExpander_SingleJob.xaml.cs
+++ b/RoboSharp.BackupApp/JobHistoryExpander_SingleJob.xaml.cs
@@ -76,13 +76,14 @@ namespace RoboSharp.BackupApp
             ByteStat = list.BytesStatistic;
             DirStat = list.DirectoriesStatistic;
             FileStat = list.FilesStatistic;
+            
+            IsResultsListBound = true;
+
+            ShowResultsListSummary(null, new System.ComponentModel.PropertyChangedEventArgs(""));
             DirectoriesStatistic_PropertyChanged(null, null);
             FilesStatistic_PropertyChanged(null, null);
             BytesStatistic_PropertyChanged(null, null);
             
-            
-            IsResultsListBound = true;
-
             ////Trigger List Update
             DirStat.PropertyChanged += ShowResultsListSummary;
             FileStat.PropertyChanged += ShowResultsListSummary;
--- a/RoboSharp.BackupApp/JobHistoryExpander_SingleJob.xaml.cs
+++ b/RoboSharp.BackupApp/JobHistoryExpander_SingleJob.xaml.cs
@@ -121,7 +121,7 @@ namespace RoboSharp.BackupApp
                 () =>
                 {
                     lbl.Content = stat?.ToString(false, true, "\n", false) ?? "";
-                    if (!IsResultsListBound) ShowSelectedJobSummary(); else ShowResultsListSummary(null, null);
+                    if (!IsResultsListBound) ShowSelectedJobSummary();
                 });
         }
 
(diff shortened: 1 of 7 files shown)


PCAssistSoftware:
Confirmed that this resolves issue from this morning.

BUT noticed another issue and not sure if related so will await your response before merging

After running a couple of single jobs the overall results 'summary' is empty 

image

RFBomb:
EDIT: Moved to first post, since this explains the issue corrected.

PCAssistSoftware:
@RFBomb - further to my last post above about missing 'overall results - summary

I am right in saying that this now shows differently than before - as 'overall results - summary' now showing source: and destination: which is not relevant for overall - and no longer showing exit code or any jobs cancelled which it did show before?

image

RFBomb:
I have that fixed locall, I ran into a race condition I'm fixing.

PCAssistSoftware:
Do you want me to merge this one to fix #120 issue or wait for you to make additional changes?

RFBomb:
Additional Changes incoming

RFBomb:
> I am right in saying that this now shows differently than before - as 'overall results - summary' now showing source: and destination: which is not relevant for overall - and no longer showing exit code or any jobs cancelled which it did show before?

I believe what you are referring to is a quirk of the code. In my opinion its a non-issue. Once a result has been added to the list, it shows properly. Its just that prior to any result being added to the list, it shows the default value.

PCAssistSoftware:
That has resolved the issue I had above - are you happy for me to now merge?

PCAssistSoftware:
> > I am right in saying that this now shows differently than before - as 'overall results - summary' now showing source: and destination: which is not relevant for overall - and no longer showing exit code or any jobs cancelled which it did show before?
> 
> I believe what you are referring to is a quirk of the code. In my opinion its a non-issue. Once a result has been added to the list, it shows properly. Its just that prior to any result being added to the list, it shows the default value.

Thanks for explanation, agreed - fine with me

RFBomb:
> Thanks for explanation, agreed - fine with me

I had a boolean being set in the wrong location in the code. (Set after the string was created instead of before). Resolved.

RFBomb:
Ok, this should be mergable now. I think that covers all the things in #120 and based on my testing, its accurately getting the results again. (I had 240 small files being copied on an NVME drive, so the copy operation was finishing before all the log lines were being added to ResultsBuilder since the UI was updating every time a log line was added due to ProgressEstimator's updating of the stats. 

ProgressEstimator now pushes updates every 125ms for the duration of its life, until the internal 'GetResults' method is called.

PCAssistSoftware:
Just tested here on multiple different sources and all okay, so will merge now

PR #143, opened 2022-01-19
Minor Fixes

RFBomb:
RoboCommand
- Fixed RoboCommand.Clone not including JobOptions
- RoboCommand now throws is starting a process while that command already has a process running.
- Insert WorkAround to allow saving of a RoboCopyJob while its running - this is done by cloning the command, then disposing of the clone.

SelectionOptions
- Improve ObsoleteTags to be more descriptive.
- Imrpove and test Regex.
- Hide Compiler warning in MainWindow


PCAssistSoftware:
ExcludeFiles / ExcludeDirectories now seems to be working fine as per yesterday's testing and comments :)

Save Job works, but when then using Load Job it is doesn't seem to process the exclusions even though showing in RCJ file - they don't show in GUI -OR- in CommandOptions etc

RFBomb:
a bool was swapped due to copy paste, this commit I just pushed seems to have resolved it for me. 
XD and XF are now importing properly from job files that are exported.

PCAssistSoftware:
> XD and XF are now importing properly from job files that are exported.

Confirmed fully working for me too

Happy to merge if you are ready?

PCAssistSoftware:
Also fully tested with SelectionOptions.ExcludedFiles.**AddRange** and SelectionOptions.ExcludedFiles.**Add**

RFBomb:
yea its good

PR #145, opened 2022-01-19
Overridable RoboCommand Properties and Methods

RFBomb:
These changes are to more easily accomodate using classes derived from RoboComand with the RoboQueue Object.

- Mark most Properties as abstract to allow consumers to override them if needed
- Mark the 'Start' methods in RoboCommand as virtual.
- Introduce 'ListOnly' start methods to RoboCommand
- RoboQueue updated to use more efficient ListOnly start method

### DIFF
--- a/RoboSharp.BackupApp/MainWindow.xaml
+++ b/RoboSharp.BackupApp/MainWindow.xaml
@@ -208,6 +208,7 @@
                                             <RowDefinition />
                                             <RowDefinition />
                                             <RowDefinition />
+                                            <RowDefinition />
                                         </Grid.RowDefinitions>
                                         <Grid.ColumnDefinitions>
                                             <ColumnDefinition Width="Auto" />
@@ -218,12 +219,14 @@
                                         <TextBlock Grid.Row="0" Grid.Column="0" Text="Verbose Output (/V)" />
                                         <TextBlock Grid.Row="1" Grid.Column="0" Text="No File Size (/NS)" />
                                         <TextBlock Grid.Row="2" Grid.Column="0" Text="No Progress (/NP)" />
+                                        <TextBlock Grid.Row="3" Grid.Column="0" Text="List Only (/L)" />
 
                                         <GridSplitter Width="5" Grid.Row="0" Grid.RowSpan="100" Grid.Column="1" ResizeBehavior="PreviousAndNext" ResizeDirection="Columns" />
 
                                         <CheckBox Grid.Row="0" Grid.Column="2" x:Name="VerboseOutput" Margin="10,0,0,0" IsChecked="True" />
                                         <CheckBox Grid.Row="1" Grid.Column="2" x:Name="NoFileSizes" Margin="10,0,0,0" />
                                         <CheckBox Grid.Row="2" Grid.Column="2" x:Name="NoProgress" Margin="10,0,0,0" />
+                                        <CheckBox Grid.Row="3" Grid.Column="2" x:Name="ChkListOnly" Margin="10,0,0,0" />
 
                                     </Grid>
                                 </Expander>
--- a/RoboSharp.BackupApp/MainWindow.xaml
+++ b/RoboSharp.BackupApp/MainWindow.xaml
@@ -147,6 +147,7 @@
                                             <RowDefinition />
                                             <RowDefinition />
                                             <RowDefinition />
+                                            <RowDefinition />
                                         </Grid.RowDefinitions>
                                         <Grid.ColumnDefinitions>
                                             <ColumnDefinition Width="Auto" />
@@ -161,7 +162,8 @@
                                         <TextBlock Grid.Row="4" Grid.Column="0" Text="Exclude Files (/XF File File ...)" />
                                         <TextBlock Grid.Row="5" Grid.Column="0" Text="Exclude Directories (/XD Dir Dir ...)" />
                                         <TextBlock Grid.Row="6" Grid.Column="0" Text="Exclude Older (/XO)" />
-                                        <TextBlock Grid.Row="7" Grid.Column="0" Text="Exclude Junction Points (/XJ)" />
+                                        <TextBlock Grid.Row="7" Grid.Column="0" Text="Exclude NEWER (/XN)" />
+                                        <TextBlock Grid.Row="8" Grid.Column="0" Text="Exclude Junction Points (/XJ)" />
 
                                         <GridSplitter Width="5" Grid.Row="0" Grid.RowSpan="100" Grid.Column="1" ResizeBehavior="PreviousAndNext" ResizeDirection="Columns" />
 
@@ -172,7 +174,8 @@
                                         <TextBox Grid.Row="4" Grid.Column="2" x:Name="ExcludeFiles" Margin="10,0,0,0" ToolTip="Separate entries with space or comma&#x0a;If filename or path contains spaces then enclose in quotes&#x0a;e.g. &quot;my file.txt&quot;,file1.txt file2.txt"/>
                                         <TextBox Grid.Row="5" Grid.Column="2" x:Name="ExcludeDirectories" Margin="10,0,0,0" ToolTip="Separate entries with space or comma&#x0a;If filename or path contains spaces then enclose in quotes&#x0a;e.g. &quot;my file.txt&quot;,file1.txt file2.txt"/>
                                         <CheckBox Grid.Row="6" Grid.Column="2" x:Name="ExcludeOlder" Margin="10,0,0,0" />
-                                        <CheckBox Grid.Row="7" Grid.Column="2" x:Name="ExcludeJunctionPoints" Margin="10,0,0,0" />
+                                        <CheckBox Grid.Row="7" Grid.Column="2" x:Name="ExcludeNewer" Margin="10,0,0,0" />
+                                        <CheckBox Grid.Row="8" Grid.Column="2" x:Name="ExcludeJunctionPoints" Margin="10,0,0,0" />
 
                                     </Grid>
                                 </Expander>
--- a/RoboSharp.BackupApp/MainWindow.xaml.cs
+++ b/RoboSharp.BackupApp/MainWindow.xaml.cs
@@ -187,6 +187,7 @@ namespace RoboSharp.BackupApp
             copy.LoggingOptions.VerboseOutput = VerboseOutput.IsChecked ?? false;
             copy.LoggingOptions.NoFileSizes = NoFileSizes.IsChecked ?? false;
             copy.LoggingOptions.NoProgress = NoProgress.IsChecked ?? false;
+            copy.LoggingOptions.ListOnly = ChkListOnly.IsChecked ?? false;
             copy.StopIfDisposing = true;
             return copy;
         }
@@ -254,6 +255,7 @@ namespace RoboSharp.BackupApp
             VerboseOutput.IsChecked = copy.LoggingOptions.VerboseOutput;
             NoFileSizes.IsChecked = copy.LoggingOptions.NoFileSizes;
             NoProgress.IsChecked = copy.LoggingOptions.NoProgress;
+            ChkListOnly.IsChecked = copy.LoggingOptions.ListOnly;
         }
 
         void DebugMessage(object sender, Debugger.DebugMessageArgs e)
--- a/RoboSharp.BackupApp/MainWindow.xaml.cs
+++ b/RoboSharp.BackupApp/MainWindow.xaml.cs
@@ -175,6 +175,7 @@ namespace RoboSharp.BackupApp
 #pragma warning restore CS0618 
             copy.SelectionOptions.ExcludeOlder = ExcludeOlder.IsChecked ?? false;
             copy.SelectionOptions.ExcludeJunctionPoints = ExcludeJunctionPoints.IsChecked ?? false;
+            copy.SelectionOptions.ExcludeNewer = ExcludeNewer.IsChecked ?? false;
 
             // retry options
             if (!string.IsNullOrWhiteSpace(RetryCount.Text))
@@ -242,6 +243,8 @@ namespace RoboSharp.BackupApp
 #pragma warning restore CS0618
             ExcludeOlder.IsChecked = copy.SelectionOptions.ExcludeOlder;
             ExcludeJunctionPoints.IsChecked = copy.SelectionOptions.ExcludeJunctionPoints;
+            ExcludeNewer.IsChecked = copy.SelectionOptions.ExcludeNewer;
+
 
             // retry options
             RetryCount.Text = copy.RetryOptions.RetryCount.ToString();
--- a/RoboSharp/CopyOptions.cs
+++ b/RoboSharp/CopyOptions.cs
@@ -126,15 +126,15 @@ namespace RoboSharp
         /// <summary>
         /// The source file path where the RoboCommand is copying files from.
         /// </summary>
+        public virtual string Source { get { return _source; } set { _source = value.CleanDirectoryPath(); } }
         private string _source;
-        /// <inheritdoc cref="_source"/>
-        public string Source { get { return _source; } set { _source = value.CleanDirectoryPath(); } }
-        /// <summary>
-        /// The destination file path where the RoboCommand is copying files to.
+
+        /// <summary> 
+        /// The destination file path where the RoboCommand is copying files to. 
         /// </summary>
+        public virtual string Destination { get { return _destination; } set { _destination = value.CleanDirectoryPath(); } }
         private string _destination;
-        /// <inheritdoc cref="_destination"/>
-        public string Destination { get { return _destination; } set { _destination = value.CleanDirectoryPath(); } }
+
         /// <summary>
         /// Allows you to supply a set of files to copy or use wildcard characters (* or ?). <br/>
         /// JobOptions file saves these into the /IF (Include Files) section
@@ -154,43 +154,43 @@ namespace RoboSharp
         /// Copies subdirectories. Note that this option excludes empty directories.
         /// [/S]
         /// </summary>
-        public bool CopySubdirectories { get; set; }
+        public virtual bool CopySubdirectories { get; set; }
         /// <summary>
         /// Copies subdirectories. Note that this option includes empty directories.
         /// [/E]
         /// </summary>
-        public bool CopySubdirectoriesIncludingEmpty { get; set; }
+        public virtual bool CopySubdirectoriesIncludingEmpty { get; set; }
         /// <summary>
         /// Copies only the top N levels of the source directory tree. The default is
         /// zero which does not limit the depth.
         /// [/LEV:N]
         /// </summary>
-        public int Depth { get; set; }
+        public virtual int Depth { get; set; }
         /// <summary>
         /// Copies files in Restart mode.
         /// [/Z]
         /// </summary>
-        public bool EnableRestartMode { get; set; }
+        public virtual bool EnableRestartMode { get; set; }
         /// <summary>
         /// Copies files in Backup mode.
         /// [/B]
         /// </summary>
-        public bool EnableBackupMode { get; set; }
+        public virtual bool EnableBackupMode { get; set; }
         /// <summary>
         /// Uses Restart mode. If access is denied, this option uses Backup mode.
         /// [/ZB]
         /// </summary>
-        public bool EnableRestartModeWithBackupFallback { get; set; }
+        public virtual bool EnableRestartModeWithBackupFallback { get; set; }
         /// <summary>
         /// Copy using unbuffered I/O (recommended for large files).
         /// [/J]
         /// </summary>
-        public bool UseUnbufferedIo { get; set; }
+        public virtual bool UseUnbufferedIo { get; set; }
         /// <summary>
         /// Copies all encrypted files in EFS RAW mode.
         /// [/EFSRAW]
         /// </summary>
-        public bool EnableEfsRawMode { get; set; }
+        public virtual bool EnableEfsRawMode { get; set; }
         /// <summary>
         /// This property should be set to a string consisting of all the flags to include (eg. DAT; DATSOU)
         /// Specifies the file properties to be copied. The following are the valid values for this option:
@@ -215,47 +215,47 @@ namespace RoboSharp
         /// Copies files with security (equivalent to /copy:DAT).
         /// [/SEC]
         /// </summary>
-        public bool CopyFilesWithSecurity { get; set; }
+        public virtual bool CopyFilesWithSecurity { get; set; }
         /// <summary>
         /// Copies all file information (equivalent to /copy:DATSOU).
         /// [/COPYALL]
         /// </summary>
-        public bool CopyAll { get; set; }
+        public virtual bool CopyAll { get; set; }
         /// <summary>
         /// Copies no file information (useful with Purge option).
         /// [/NOCOPY]
         /// </summary>
-        public bool RemoveFileInformation { get; set; }
+        public virtual bool RemoveFileInformation { get; set; }
         /// <summary>
         /// Fixes file security on all files, even skipped ones.
         /// [/SECFIX]
         /// </summary>
-        public bool FixFileSecurityOnAllFiles { get; set; }
+        public virtual bool FixFileSecurityOnAllFiles { get; set; }
         /// <summary>
         /// Fixes file times on all files, even skipped ones.
         /// [/TIMFIX]
         /// </summary>
-        public bool FixFileTimesOnAllFiles { get; set; }
+        public virtual bool FixFileTimesOnAllFiles { get; set; }
         /// <summary>
         /// Deletes destination files and directories that no longer exist in the source.
         /// [/PURGE]
         /// </summary>
-        public bool Purge { get; set; }
+        public virtual bool Purge { get; set; }
         /// <summary>
         /// Mirrors a directory tree (equivalent to CopySubdirectoriesIncludingEmpty plus Purge).
         /// [/MIR]
         /// </summary>
-        public bool Mirror { get; set; }
+        public virtual bool Mirror { get; set; }
         /// <summary>
         /// Moves files, and deletes them from the source after they are copied.
         /// [/MOV]
         /// </summary>
-        public bool MoveFiles { get; set; }
+        public virtual bool MoveFiles { get; set; }
         /// <summary>
         /// Moves files and directories, and deletes them from the source after they are copied.
         /// [/MOVE]
         /// </summary>
-        public bool MoveFilesAndDirectories { get; set; }
+        public virtual bool MoveFilesAndDirectories { get; set; }
         /// <summary>
         /// This property should be set to a string consisting of all the attributes to add (eg. AH; RASHCNET).
         /// Adds the specified attributes to copied files.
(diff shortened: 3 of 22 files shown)


PCAssistSoftware:
Ready to merge?

RFBomb:
Not Quite, I think its pretty done, but I'm still reworking my application to use RoboQueue.

I previously had a fairly complicated class wrapping RoboCommand, that is not derived instead of a wrapper. to be able to use it with RoboQueue. Once I perform some copy operations in my application I'll let you know.

PCAssistSoftware:
okay, no problem

PCAssistSoftware:
Thank you :)

Tested on excluding both files and directories and appears to be working fine!

PCAssistSoftware:
@RFBomb

Question for you.

Is there any way of knowing / retreiving during RoboQueue events such as OnFileProcessed, OnCommandCompleted, OnCopyProgressChanged etc the currently running Job Name?  The reason I ask is that in my tool I am creating I will be updating results live / realtime in a grid control (provided by DevExpress) so need to know the name of the currently running Job to be able to update the correct row in my grid?   Thanks

RFBomb:
The `sender` should be the RoboCommand object that raised the event. RoboQueue should be acting as a simple pass-through here. Its even using the same delegate as RoboCommand.

PCAssistSoftware:
> The `sender` should be the RoboCommand object that raised the event. RoboQueue should be acting as a simple pass-through here. Its even using the same delegate as RoboCommand.

Thanks, I was looking for it under `e` rather than `sender` :)

RFBomb:
Ok, 
I thought it was good since it was working fine in the backup-app, but apparently this needed quite a bit of work for consumers. The good news here is that this is now working with my app. 

I was able to start/stop/pause/resume/etc properly, and it successfully worked fine with my 18 commands running all at once.

RFBomb:
Ok, as long as this works with your application (and no other issues are seen that this doesn't already fix between now and merge), I think this is fine to merge in as far as I can tell.

PCAssistSoftware:
Just tested it with my new app and also with my original GUI I have been using for years and it all seems to work as expected, tried list only, cancel, stop and all the other usual combinations and all files copied as expected and all information showing in GUI as expected.  Therefore if you are happy I will merge it?

RFBomb:
yea, thats fine. 

Also, I was able to duplicate the issue with the 'live results' from the estimators not being displayed in the backup app. Not entirely sure whats going on there yet, but its a 'cosmetic' issue (that usually only occurs when stepping through the code, specifically stepping on anything ProgressEstimator related). But no matter what, even when that occurs, the final results and once all threads ran, everything added up correctly. So I think leave that to another day. (I think it has something to do with not being subscribed to properly in the backupapp, or the task to perform the count not starting when RoboQueue starts, unsure.)

This thing has enough fixes I think are important to merge to let that one thing slide until we have time to track it down. The important bits are all working smoothly as far as I can tell.

PCAssistSoftware:
> Also, I was able to duplicate the issue with the 'live results' from the estimators not being displayed in the backup app. Not entirely sure whats going on there yet, but its a 'cosmetic' issue (that usually only occurs when stepping through the code, specifically stepping on anything ProgressEstimator related). But no matter what, even when that occurs, the final results and once all threads ran, everything added up correctly. So I think leave that to another day. (I think it has something to do with not being subscribed to properly in the backupapp, or the task to perform the count not starting when RoboQueue starts, unsure.)

That's fine - just glad you managed to duplicate it and it wasn't just me going crazy...

RFBomb:
Yea for some reason, something that changed in this PR made it near 100% reproducible. 

I'm noticing it (almost every time, breakpoint or not) specifically on single job tab, I haven't seen it occur at all on the multi job tab. 

I think it has to do with the event listener that writes the progress estimator values.

PCAssistSoftware:
Cancelling a job is not reflected in ExitCode for multi-job

For single-job it correctly shows ExitCode: 0 (NoErrorNoCopy) but it seems for multi-job it shows ExitCode 2 (ExtraFilesOrDirectoriesDetected) or similar both in summary and results grid

RFBomb:
Multi-job is unique because of how jobs statuses are combined. 

0 - NoErrorNoCopy is unable to combine with any other status. Same for Canceled ( -1 ). 

So other properties have been added to show if any jobs were cancelled or no error no copy.

PCAssistSoftware:
I am looking at the individual results rather than combined results e.g. in RoboQueue_OnCommandCompleted I see the results for each command as completed

RFBomb:
That confuses me, since it's literally just awaiting the same task as single job does. RoboQueue is just starting them up in sequence, and awaiting them. RoboCommand still uses all the same logic for deliver each commands results.

RFBomb:
There's some post processing to decide which results list the results are added to, but RoboQueue.OnCommandCompleted passes along the original args generated by the command

PCAssistSoftware:
Thats what I thought, and a single job returns the correct exit code if cancelled, but when cancelling a multi-job the individual results are showing wrong exit codes, will test more tomorrow and come back with more info, only noticed it as am adding results from each job into a grid as they are returned and thought it was a bit odd

RFBomb:
Yea, that confuses me since it's still issuing the 'RoboCommand.Stop' to each command, so the act of cancelling should act the same way no matter what.

PCAssistSoftware:
So add a couple of jobs to RoboQueue - check listonly - start and immediately stop it and surely you would expect exit code to be ExitCode: 0 (NoErrorNoCopy) - but not it is ExitCode 1 (FilesCopiedSuccessfully) - and that quick test was in BackupApp not my own code

RFBomb:
I think I know what’s happening.
It sounds like it’s using the results status from ProgressEstimator and not
the process. ProgressEstimator has no way to know if the job was cancelled,
so it uses the file & dir counts.

RFBomb:
Results generated here:
https://github.com/tjscience/RoboSharp/blob/52d11e76084a7732996140bebfcdef8964fcfc72/RoboSharp/RoboCommand.cs#L527

which calls this method, and results from ProgressEstimator are replaced with the results from Process.ExitCode as defined above:
https://github.com/tjscience/RoboSharp/blob/52d11e76084a7732996140bebfcdef8964fcfc72/RoboSharp/Results/ResultsBuilder.cs#L89

I would breakpoint the line 89 in the ResultsBuilder and see what its doing when it occurs. Because thats the only time the RoboCopyResults.Status where all RoboCommand results get their status.

PCAssistSoftware:
Okay so debugging at the line suggested has helped narrow down the issue and the job status is only wrong when using listonly

When NOT using listonly and stopping the job you correctly get

image

But when using listonly and stopping you get

image

RFBomb:
What was the value of the exit code int variable though? That's the deciding factor, since that's what's being returned by the process

PCAssistSoftware:
> What was the value of the exit code int variable though? That's the deciding factor, since that's what's being returned by the process

exitCode = 1

RFBomb:
I'll take a look, but if that's what's being submitted to the method from the RoboCommand task when results are resquested, then it's what's being reported from the process. 

Only other option is to force a value into the call for in the case of cancellation requested. ( maybe by looking at the 'wasCancelled' set inside the stop method)

PCAssistSoftware:
Works fine on single jobs

PCAssistSoftware:
EDIT: hold fire for now, seems to potentially be related to sources not containing much data - perhaps it is completing so quickly that I am not stopping it quickly enough even though GUI shows stopped, the background RoboCopy task has still returned as if completed - will continue testing and get back to you :)

RFBomb:
Yea, that's a race condition we can't solve lol

It will always take time for the processor to loop through everything RoboQueue does to issue the stop command and react, that's why I had to put the additional checks into Stop/resume/suspend methods. Because there was potential to try killing/pausing/resuming a process that already exited.

All my tests have several subdirectories and like 200 files in it. 

Edit: for list only tests, I typically use folders on two separate HDD drives. That way the list operation is slow because two disks have to spin up, instead of using SSDs.

PCAssistSoftware:
Good idea, I had made mistake of using small folders on my desktop, so will do as suggested from now on - thanks for bearing with me

RFBomb:
The nice thing here is that you can now use the 'save /load job buttons and save these tests to a folder. 
Really wish that was put that in before building RoboQueue lol. Would be made the testing less hassle.

PCAssistSoftware:
Very true, constantly entering paths and choosing options in the GUI was time consuming, in the end whilst I was testing I just hardcoded them in to app temporarily.

PCAssistSoftware:
Confirmed with multiple test that the issues I was having was definitely due to using small sources, all fine now.

Separate point - overall speeds seem to be showing rounded down with no decimal points - pretty certain we had discussed this at some point and you had set it to 2 or 3 decimal places same as it is for each jobs results?

image

Adding all the new features to a new GUI just for fun has been a great way to fully test everything and to fully appreciate all the work you had done :)  I have had a couple of issues but want to fully check them out myself before I comment further just in case my coding

RFBomb:
Line 340 in speed stats file:

`long i = Divisor < 1 ? 0 : (long)Math.Round(Combined_BytesPerSec / Divisor, 3);`

Type of `i` should be double or var, and cast be removed. 

The next rounding statement In the method also needs `(long)` cast removed.

PCAssistSoftware:
Thanks

Changed to var and now showing 3 decimal places but confused re calculation

Job 1 was 111.385 mb/min and Job 2 was 128.926

So added together and divided by number of jobs - 113.385 + 128.926 = 240.311 / 2 = 120.155

But GUI shows 117.232

RFBomb:
Divisor in object is likely calculated to 3

PCAssistSoftware:
> Divisor in object is likely calculated to 3

Confused as if divisior was three then result shown would be 80.103

RFBomb:
Idk then, that was just the first thought that came to mind, I didn’t
actually do math or check anything

PCAssistSoftware:
Okay, it was certainly working at one point i'm sure, as remember discussing it with you and checking the results.

Just tested it under single-job as wel - ran two jobs and the overall seems to just show same values as results from second job for speed stat average - all other values fine.

PCAssistSoftware:
Question - Progress Estimator - just been testing on some much larger sources e.g. 80000+ files etc - when we first started testing am I correct in saying that the progress estimator updated results in the GUI regularly? which is what I would expect if it is showing estimate of progress? if so then when running larger jobs where it is more noticeable then it doesn't actually update until job has finished, at which point you may as well look at results under job history or list-only / run results?

This test is being done in the BackupApp but can replicate it in my own test GUI as well?

Thoughts, or am I just going mad thinking it was showing live updates as job progressed?

This video clip is good example of what I mean - https://www.screencast.com/t/KQgXPiI5DU

PCAssistSoftware:
Okay I have spent some time this evening debugging the average statistics issue, and hoping you might be able to throw some light on the results I am seeing.

I am running two jobs - and looking for now just at MegaBytesPerMin as I think when we solve one the solution will solve BytesPerSec.

Looking at the results for each of the two jobs individually both in GUI and values in code:

```
job 1 mb/min		48.332
job 2 mb/min		46.261
```

Looking at values of Combined_MegaBytesPerMin and Divisor whilst GUI is running - I was outputting values to console and it seems to calculate three times rather than the expected two

```
**Combined_MegaBytesPerMin: 48.332
**Divisor: 1
**Combined_MegaBytesPerMin: 96.664
**Divisor: 2
**Combined_MegaBytesPerMin: 142.925
**Divisor: 3
```

So the first value (48.332) is correct and tallies above, then it seems to add the value from job 1 again (48.332 + 48.332 = 96.664), then it adds the value from job 2 (48.332 + 48.332 + 46.261 = 142.925)  and then divides by 3 and returns obviously wrong value.

So why is it adding job 1 values TWICE? as if it didn't do that then figures would be correct, divisor would then be 2 rather than 3 and problem solved - just can't figure out yet why it is adding first job results twice?

Any clues as to where I can look would be appreciated - thanks

PCAssistSoftware:
debugging further

When first job is running it is calling `internal void CalculateAverage()` on line 336 of SpeedStatistic.cs TWICE:-

Firstly from line 371 of SpeedStatistics:-

```
public void Average(IEnumerable<ISpeedStatistic> stats)
        {
            Add(stats);
            CalculateAverage();
        }
```

and then from line 291 of RoboCopyResultsList.cs:-

```
 if (Average_SpeedStatsField.IsValueCreated)
                    {
                        Average_SpeedStatsField.Value.Add(r?.SpeedStatistic);
                        if (RaiseValueChangeEvent) Average_SpeedStatsField.Value.CalculateAverage();
                    }
```


But when second job runs it only calls `internal void CalculateAverage()` ONCE from line 291 of RoboCopyResultsList.cs

Hoping this might mean something to you as to why calling it twice for first job?

RFBomb:
Thanks for taking a look at this. I can figure it out from here now that I know exactly what its doing.
I just found an issue where progressestimator.CurrentFile == null and data = 100% is being reported. 
Mostly happens on Thumbs.db, but happens elsewhere too. I've no clue why, it seems like robocopy is reporting it multiple events multiple times sometimes.

RFBomb:
Ok, so the issue I found was that sometimes RoboCopy repeats log items!
Specifically, RoboCopy issues the '100%' token multiple times in a row on some files. This took a while to track down. but its the cause of the error I was chasing. 

image


The non-updating issue progressestimator values was do to a badly formatteted DoWhile loop. I've updated it to always do, then check the if more time is available (previously, it checked time avialable to do before, which meant it always returned false from the method, which meant it never pushed an update)

PCAssistSoftware:
Well spotted on both points :)

RFBomb:
I'm having a nightmare of a time dealing with troubleshooting the speed thing.
I'm not getting the speed results at all, seems to be the loglines not showing it. in the past hour, i've run maybe 15 tests, and only got speed results once. Various file sizes, folder counts, etc. 

ProgressEstimator may be part of the culprit here, but I'm not sure. The RoboCommand task that is requesting the results seems to be requesting the results before all of the events propogate. Isn't ASYNC programming fun?

I don't know how to syncronize this.

PCAssistSoftware:
That's not good. I assume must be related to a recent change you have made as all tests I have run today and that is a lot were all returning loglines okay.

Let me know if anything I can do or test for you

RFBomb:
My cpu usage was also hitting 100% while debugging this. My work laptop been struggling lol. 

Now that i know what some of the issues were, I plan to make another branch and Refix the issues, discarding all my testing changes. Hopefully that resolves some of it.

I'm wondering if there's a way to see if events are waiting to be raised, and wait for them to propagate, for slower CPUs that may bottleneck like mine seems to

PCAssistSoftware:
I had the 100% issue earlier, was one of the issues I was alluding too that I was going to come back to later as wasn't sure if it was just my app though as been making so many changes.

Okay, here to test whatever you do.

I am by no means an expert at Async, it is great when it works but I know from other tools I have written can be a total pain when one event relies on another finishing etc.  But your logic makes sense and you would think there would be some way to do that.

RFBomb:
Ok, so after some more reading on threading, I'm gonna make some changes to it. 

Apparently tasks marked with LongRunning always create a new thread. 

Use of Await unschedules it off the thread, then when it resumes it requests access again, and may be put into a different thread. So making RoboCommand.Start Async and long running is counter-productive, since it introduce more (albeit not much more) work. 

The real ticker here is using that pattern for progress-estimator and RoboQueue I think. All the await Task.Delays that occur introduce context and thread switching (since the thread is released and must be re-requested). This might be where the high CPU usage is coming in. 

So I'll re-evaluate these and see what can be done. I also want to figure out which thread the RoboCommand events are firing on.

PCAssistSoftware:
Okay that make perfect sense, look forward to testing whatever you come up with :)

PCAssistSoftware:
> The `sender` should be the RoboCommand object that raised the event. RoboQueue should be acting as a simple pass-through here. Its even using the same delegate as RoboCommand.

For info in some events sender.Name is blank and not showing the current Job Name? Not sure if it is possible to fix? 

image

This applies to `RoboQueue_OnCommandStarted`, `RoboQueue_ListResultsUpdated` and  `RoboQueue_RunOperationResultsUpdated`

For RoboQueue_OnCommandStarted I can get it through e.Command.Name but it would be nice to be consistent and always get it through sender.Name

`RoboQueue_OnFileProcessed`, `RoboQueue_OnCommandCompleted`, `RoboQueue_OnCopyProgressChanged` are all working fine

Thanks

PCAssistSoftware:
If I cancel a job then as expected WasCancelled is True, but Successful is also True? surely if cancelled then Successful should be False?

image

RFBomb:
Easy fix for that bool check -> it's checking if the int is less than 0x10, but. Typically this would be 0-10. But -1 also works. So the bool check should be 


> public bool Successful => !WasCancelled && ExitCodeValue < 0x10;

PCAssistSoftware:
> Easy fix for that bool check -> it's checking if the int is less than 0x10, but. Typically this would be 0-10. But -1 also works. So the bool check should be
> 
> > public bool Successful => !WasCancelled && ExitCodeValue < 0x10;

Excellent - thank you

PCAssistSoftware:
@RFBomb - advice/help please

Been testing for a few hours and everything seems to be working as expected in BackupApp which is great.

I do however have odd issue with my own testing app where it seems to freeze for ages before updating the interface when the source is large - for example I am using C:\ as source and just running ListOnly for this test.

I have spent the morning recreating my test app adding each part a step at a time to determine what area is causing the issue and I have narrowed it down to the OnFileProcessed event - if I remove the handler or comment out the code in that section then it runs perfectly fine and estimator updates results quickly as expected, but if I add the handler and have code which simply shows current operation, filename, and size then app freezes for ages and then suddenly comes back to life with estimator complete.

First video clip shows it running with OnFileProcessed commented out - https://www.screencast.com/t/gJEbffdTbm0P

Second clip shows it with OnFileProcessed NOT commented out - https://www.screencast.com/t/oAwRw8SyWE - clip trimmed as this freeze goes on for several minutes before it then shows estimator results OR shows a context deadlock error message

Attached is an extremely basic example app (in VB) - just needs destination path changing to suit your setup - try running it first with line 40 commented out and then without line 40 commented out.

Thanks

[roboqueue_performance_issue.zip](https://github.com/tjscience/RoboSharp/files/7977819/roboqueue_performance_issue.zip)

RFBomb:
Can you commit that bool update directly?
Also, the reason you aren't seeing the JOB NAME in the event's sender is because its ROBOQUEUE sending the event, not ROBOCOMMAND. The RoboQueue is starting some RoboCommand, hence why the command is in the args, sender is RQ. 

As for your project: You are just overloading the UI thread with event dispatches.
Updating all the labels like that requires each one to be invoked, so now all the events are fighting for control of the UI. 
It doesn't happen with line 40 commented out because you only have 3 events fighting, and they are fired one after the other, so not fight actually occurs. 

But since RoboCommand is on a seperate thread than RoboQueue, now you have 2 threads fighting for control, one is sending hundreds of requests, while the other updates every 60-100ms or so. So you the UI thread is constantly processing the fileProcessed stuff, without ever giving PE a chance to complete its event.

SyncLock is added to ensure multiple threads aren't fighting. I had to add it to RQ and PE in RoboSharp to combat the same thing.

[roboqueue_performance_issue.zip](https://github.com/tjscience/RoboSharp/files/7978413/roboqueue_performance_issue.zip)

PCAssistSoftware:
> Can you commit that bool update directly? 

Should already have been done earlier

PCAssistSoftware:
> Also, the reason you aren't seeing the JOB NAME in the event's sender is because its ROBOQUEUE sending the event, not ROBOCOMMAND. The RoboQueue is starting some RoboCommand, hence why the command is in the args, sender is RQ.

Okay that makes sense, so is there a way I CAN see it somehow in those events?

> As for your project: You are just overloading the UI thread with event dispatches. Updating all the labels like that requires each one to be invoked, so now all the events are fighting for control of the UI. It doesn't happen with line 40 commented out because you only have 3 events fighting, and they are fired one after the other, so not fight actually occurs.
> 
> But since RoboCommand is on a seperate thread than RoboQueue, now you have 2 threads fighting for control, one is sending hundreds of requests, while the other updates every 60-100ms or so. So you the UI thread is constantly processing the fileProcessed stuff, without ever giving PE a chance to complete its event.
> 
> SyncLock is added to ensure multiple threads aren't fighting. I had to add it to RQ and PE in RoboSharp to combat the same thing.
> 

Excellent - thank you for that explanation.  Now updating correctly for curent filename, size, operation but the estimator now updates much slower - what is best way to get that updating as it was before?

RFBomb:
FYI, the SyncLock and the update UI tasks in both ProgressEstimator objects was put in their to essentially prevent that very thing (UI event Overload) from occurring when many file processed events occur in rapid succession.

RFBomb:
> what is best way to get that updating as it was before

That's only updating slowly due to the SyncLock. Only way around it is performing the SyncLock fewer times in OnFileProcessed. 

I left a list in there with a SyncLock to showcase this. Just comment out the UI update code in the handler method. 

Best bet is only update the UI labels every X milliseconds, to avoid excessive SyncLock and UI thread overload for text that doesn't even have time to render. 

Maybe start a task that pulls in the last item from the list every X milliseconds instead of updating the UI on the event. 

(Or just use a private object to store the reference, and when the task needs to refresh, it pulls in whatever that object is. Object will have to be SyncLock for writing, not reading. 

Once this is working, maybe it can be added as an example to the wiki

RFBomb:
> Okay that makes sense, so is there a way I CAN see it somehow in those events?

I thought that the RoboCommand the event was relevant to was in the event args?

PCAssistSoftware:
> > Okay that makes sense, so is there a way I CAN see it somehow in those events?
> 
> I thought that the RoboCommand the event was relevant to was in the event args?

Perhaps just me and I am looking in wrong place (quite possible!)

In RoboQueue_OnCommandStarted I can find Name under e.Command.Name - but no sign of it under Sender as Name here is just blank

In RoboQueue_ListResultsUpdated again Sender.Name is blank and under e. all I can see is ResultsList which doesn't show me name of job which raised the event

RFBomb:
> In RoboQueue_OnCommandStarted I can find Name under e.Command.Name - but no sign of it under Sender as Name here is just blank

Did you name RoboQueue? If not, then sender won't have a name.

PCAssistSoftware:
> > what is best way to get that updating as it was before
> 
> That's only updating slowly due to the SyncLock. Only way around it is performing the SyncLock fewer times in OnFileProcessed.
> 
> I left a list in there with a SyncLock to showcase this. Just comment out the UI update code in the handler method.
> 
> Best bet is only update the UI labels every X milliseconds, to avoid excessive SyncLock and UI thread overload for text that doesn't even have time to render.
> 
> Maybe start a task that pulls in the last item from the list every X milliseconds instead of updating the UI on the event.
> 
> (Or just use a private object to store the reference, and when the task needs to refresh, it pulls in whatever that object is. Object will have to be SyncLock for writing, not reading.
> 
> Once this is working, maybe it can be added as an example to the wiki

Okay, that make sense - will need to learn about it I think as not quite sure how to implement what you say, but updating UI labels every X milliseconds make sense.  Think I just need to decide which fields are the priority e.g. current file being processed is probably more important to be up-to-date than estimator results are.

Agreed once got working example I can put on wiki - have quite a few bits here I was going to include - that was part of my reason for playing about with creating a new GUI and adding RoboSharp from scratch so I can add lots of code snippets / examples to Wiki

PCAssistSoftware:
> > In RoboQueue_OnCommandStarted I can find Name under e.Command.Name - but no sign of it under Sender as Name here is just blank
> 
> Did you name RoboQueue? If not, then sender won't have a name.

No each "job" has a name - not the queue - it is the RoboCommand job name I am after so I can update my results grid for correct job when event is raised - if that make sense

RFBomb:
>> I can find Name under e.Command.Name 

> No each "job" has a name - not the queue - it is the RoboCommand job name I am after so I can update my results grid for correct job when event is raised - if that make sense

So its working correctly. then. RoboQueue is sending the event. NOT Robocommand, which is why it works this way. Use the event args.

> In RoboQueue_ListResultsUpdated again Sender.Name is blank and under e. all I can see is ResultsList which doesn't show me name of job which raised the event

This is intended. Its just saying the collection of results changed. This includes them being emptied out.

RFBomb:
Lines 616+
image

The commands that have RoboCommand as the sender are literally being sent by the robocommands. 

Any other events are being fired by RoboQueue.

PCAssistSoftware:
Okay understood - thanks for clarification.  

I can cope with not getting job name for RoboQueue_ListResultsUpdated and RoboQueue_RunOperationResultsUpdated, and will continue to use e.Command.Name for RoboQueue_OnCommandStarted as when this event is raised I update status column in my grid to show which job(s) are running, hence needing to know name to update correct row in my grid.

Thanks

RFBomb:
OnCommandCompleted will tell you when any specific command completes

PCAssistSoftware:
> OnCommandCompleted will tell you when any specific command completes

Yes I use that to change my status field to Successful, Cancelled, Has Errors, Has Warnings etc

And I use OnCommandStarted to change my status field to Running

Thanks again for useful responses, have renamed my handlers so I easily see which ones raised by RoboCommand and which by RoboQueue - it make more sense to me now :)

PCAssistSoftware:
Definitely need to find a better solution for my GUI updates as current method completely locks the form so can't click Stop/Pause or anything

RFBomb:
For OnFileProcessed, write 'E' to an object private to the form. 

When dirs and files stats are updated update the UI with contents of E. 

That's the easiest way. Otherwise you must introduce a DateTime object, a TimeSpan object, then compare to see if you wish to do the update. (If less than timespan has elapsed, just exit the method)

The UI locking up is exactly why I had introduced the Tasks that push out deferred updates on ProgressEstimator. Otherwise every single time a file/dir is processed it would force a UI refresh due to invoking PropertyChanged. It occurred before it ever had time to render.

PCAssistSoftware:
> For OnFileProcessed, write 'E' to an object private to the form.
>
> When dirs and files stats are updated update the UI with contents of E.
> 
> That's the easiest way. Otherwise you must introduce a DateTime object, a TimeSpan object, then compare to see if you wish to do the update. (If less than timespan has elapsed, just exit the method)
> 
> The UI locking up is exactly why I had introduced the Tasks that push out deferred updates on ProgressEstimator.

Okay, not sure how to do it, but I get the logic - and will do some research as everything else now seems to work exactly as I was hoping so just GUI updates to streamline / fix.

I had one other thought and please dismiss me if not possible, but couldn't CurrentOperation, CurrentFile, and CurrentSize be returned as part of the progess estimator, then everything that needs to be shown on form in realtime (so to speak) relating to progress is all coming from one place - I may be talking rubbish here, just typing as I think really.

Thinking it through I assume I am going to have same issue when implementing CurrentFileProgress in OnCopyProgressChanged and whatever I choose to do for overall job progress as well.

Sorry just trying to learn and get this right as not issues I had previously with RoboCommand so need to re-learn best methods now using RoboQueue

RFBomb:
ProgressEstimator would be affected by this request, but RoboQueueProgressEstimator would be unable to provide those details. 

ProgressEstimator would need two more events, CurrentFileChanged and CurrentDirChanged. 

This is relatively easily doable, we just need to code it in. But remember, that would fire every time OnFileProcessed also fires. (As long as it wasnt a system message) 

CopyProgress itself is not passed into ProgressEstimator, only 'copy started (<100%) and 'copy complete (100%)' so that info would still come from RoboCommand.

PCAssistSoftware:
Not to worry if it adds more complication etc - just trying to think best way to achieve it for me and all new users coming to RoboSharp for first time creating basic apps and wanting UI updates.  

It is only really these things that are reported 'real-time' such as current file, current operation, current file size, and CurrentFileProgress for progress bar (and of course progress estimator) that are an issue, most other important information happens after jobs complete so not such a strain on the UI

RFBomb:
Personally, I like the idea of events for 'SystemMessage', 'new file' and 'new dir', as opposed to a single 'OnFileProcessed' that handles them all, even if I don't have a use for it in my application. It would it overall easier for everybody to track what happens and what to expect for different scenarios. 

If I made it up, I would have a different object for each one, just to accentuate the different types, but the current way is relatively simple, so I'm fine with it. (Now that I've put the current file and current dir into CopyProgressUpdated

PCAssistSoftware:
Sounds good to me.  

I have been playing about this afternoon and even taking everything else out of my GUI for testing and just trying to show **current file**, **current operation**, **current file size** on labels, and **CurrentFileProgress** on progress bar completely locks up my form however I do it.  

A problem I never really had so much in the past when just using RoboCommand so any solution would be gratefully received.

PCAssistSoftware:
Update: 

I have got my form responding properly - and without the need of synclock etc - the cause in my case it appears was where I had placed `queue = New RoboQueue`

Originally it was placed in Form1_Load event so was running on same thread as the main form.

Looking back at my original GUI I had created years ago for RoboCommand I realized that I needed to place it in the button click event or the Aync sub called from the button click **NOT** the Form load event and then it would run on its own thread, and work much better.

And also how I was updating the labels e.g. lblCurrentOperation.Text = e.ProcessedFile.FileClass needed to be changed so it was checking the thread and using invoke e.g. 

```
Update_TextField(lblCurrentOperation, e.ProcessedFile.FileClass)

Private Sub Update_TextField(controlname As Control, text As String)
        If controlname.InvokeRequired Then
            'We are on a secondary thread.
            controlname.Invoke(New Action(Of Control, String)(AddressOf Update_TextField), controlname, text)
        Else
            'We are on the UI thread.
            controlname.Text = text
        End If
End Sub
```

After making this change now current file, current operation, current file size labels **AND** progress estimator update fine without anything additional needed sych as synclock, and the form and all buttons are reponsive with no locking up etc.

Modified project attached for info.

[roboqueue_performance_issue_v2.zip](https://github.com/tjscience/RoboSharp/files/7986052/roboqueue_performance_issue_v2.zip)

So all in all mainly down to my coding and not thinking it through.........   though above suggestion of separate events for areas like current file, current operation, current file size and CurrentFileProgress is still probably worth a thought

PR #146, opened 2022-01-28
Thread Optimizations

RFBomb:
@PCAssistSoftware I've been working on this all day, and at the moment I don't know how else it could be improved, without someone with much more experience threading applications taking a look at it. At the moment, I've done my best on it. 

The commits all have descriptions as to what changed. Please test it out. As far as I can tell, it should now be using less threads overall, but also it seems to no longer have the issue of lines being added to the log after cancellation occurs.
Note that the UI is still running its updates after cancellation occurs, I don't think anything can be done about that. 

I've rebuilt a large amount of ProgressEstimator to accomplish this, since that has the stat objects that were all pushing events to the UI, which I believe is the cause of the issue we were seeing. This update loop I'm sure can be improved upon, but I am unsure how exactly. But overall, it is much simpler and more 'inline' with when a line is output from the process.

Some other fixes are included here too, such as apparently 'EXTRAS' are not counted towards the 'TOTAL' of a statistic object.

### DIFF
--- a/RoboSharp.BackupApp/JobHistoryExpander_SingleJob.xaml.cs
+++ b/RoboSharp.BackupApp/JobHistoryExpander_SingleJob.xaml.cs
@@ -79,16 +79,15 @@ namespace RoboSharp.BackupApp
             
             IsResultsListBound = true;
 
-            ShowResultsListSummary(null, new System.ComponentModel.PropertyChangedEventArgs(""));
+            // Initialize the values
             DirectoriesStatistic_PropertyChanged(null, null);
             FilesStatistic_PropertyChanged(null, null);
             BytesStatistic_PropertyChanged(null, null);
-            
-            ////Trigger List Update
-            DirStat.PropertyChanged += ShowResultsListSummary;
-            FileStat.PropertyChanged += ShowResultsListSummary;
-            ByteStat.PropertyChanged += ShowResultsListSummary;
 
+            //Update Summary Event
+            ShowResultsListSummary(null, new System.ComponentModel.PropertyChangedEventArgs(""));
+            list.CollectionChanged += (o,e) => ShowResultsListSummary(o, null);
+            
             ////Bind in case updates
             DirStat.PropertyChanged += DirectoriesStatistic_PropertyChanged;
             FileStat.PropertyChanged += FilesStatistic_PropertyChanged;
@@ -154,7 +153,7 @@ namespace RoboSharp.BackupApp
         /// </summary>
         private void ShowResultsListSummary(object sender, System.ComponentModel.PropertyChangedEventArgs e)
         {
-            if (e == null || (e.PropertyName != "" && e.PropertyName != "Total")) return;
+            //if (e == null || (e.PropertyName != "" && e.PropertyName != "Total")) return;
             string NL = Environment.NewLine;
             if (ResultsList == null || ResultsList.Count == 0)
             {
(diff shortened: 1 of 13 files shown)


PCAssistSoftware:
Hi,

Just ran a few tests and all seems to be working fine for me so far, no issues with CPU usage going high, and progress estimator updating in real-time.

Have also looked through all your code changes and read the descriptions - all changes made make sense to me.

Still same issue with speed stat averages - but assume you probably haven't looked at that whilst doing other bits? only asking as don't want to test and report bits which are still in progress.

If anything in particular you want me to test just let me know.

RFBomb:
I was so focused on the Process_OutputDataReceived method running after results have already been built and reworking the progress estimator to be less cpu taxing that I forgot about that completely.

PCAssistSoftware:
That's fine - it wasn't a priority compared to other issues.  Just checked speed stat issue and same as reported before, even when running just 1 job in multi-job it is adding speeds twice rather than once

I have made a couple of other points on #145 which hopefully will be easy fixes (or answers)

PCAssistSoftware:
More testing done - progress estimator is now way out for some reason - not sure how I missed when testing yesterday (sorry)

Screenshots below show differences between what progress estimator show and what run results show.

Run results are CORRECT - including EXTRAS

image


image

PCAssistSoftware:
UPDATE: Progress Estimator results are only wrong when doing COPY - they are fine with LISTONLY

image

PCAssistSoftware:
EXTRAS: in progress estimator - correct for Directories, but not for Files or Bytes - All show correctly in Run Results

Screenshots below show comparison to what I can see in log lines and what is shown in progress estimator 

image

RFBomb:
I dont know what to say, I've tried this on two seperate computers, and here is the result:

image


Everything matches up for me, in both ListOnly and standard when running single jobs

PCAssistSoftware:
The problem I reported is with **multi-job** NOT single job.

Just deleted all my source files and redownloaded clean version of PR #146 just to ensure my testing is accurate.

Add two jobs as below, and you can see that the progress estimator is showing different totals than what are showing for each of the jobs separately - works fine for ListOnly.

Hopefully image below makes it clearer.

image

RFBomb:
Found the issue, it was bad math in the statistics object. Workiung on the fix now

PCAssistSoftware:
just downloaded last commit but getting error as below?

line 548 of RoboCommand.cs

image

PCAssistSoftware:
Excellent work - I can confirm progress estimator now shows correct information and speed average also now correct.

Thanks Robert - very much appreciated.

Will test in more depth later or tomorrow, but so far so good :)

If you get a moment at some point please if you don't mind comment on my points on #145 which hopefully will be easy fixes (or answers)

PCAssistSoftware:
Just ran some more jobs of various sizes and all results for both progress estimator, list-only results and run results appearing as expected.

Let me know if you want anything specific tested and/or when ready to merge this

RFBomb:
That’s ready to merge, as it should fix all outstanding items I believe

I did create another draftPR that reworks RoboQueueProgressEstimator to use a similar method as ProgressEstimator, but am unsure which is better, hence a different PR. Both work fine. 

Essentially, current uses a task for each object and processes the events from each ProgressEstkmator RoboQueue is running in a deferred fashion, requiring three tasks running and sleeping to do so. 

The rework does all the math immediately, by using ‘Lock(Object)’ functionality, then only periodically updates the public-facing objects. This is just one task.

I reworked it to make tracking it slightly easier, since it’s explicit instead of three tasks running 1 method in parallel with each other, which is obnoxious to debug. 

But both are functional, so I made it a PR Draft so you and TJ can review.

PCAssistSoftware:
Okay I will merge this now and then I can take a look at the other PR draft to see for myself

PR #147, opened 2022-01-31
Alternative RQPE code

RFBomb:
Essentially, current uses a task for each object and processes the events from each ProgressEstkmator RoboQueue is running in a deferred fashion, requiring three tasks running and sleeping to do so.

ThIS rework does all the math immediately, by using ‘Lock(Object)’ functionality, then only periodically updates the public-facing objects. This is just one task.

I reworked it to make tracking it slightly easier, since it’s explicit instead of three tasks running 1 method in parallel with each other, which is obnoxious to debug.

But both are functional, so I made it a PR Draft so @PCAssistSoftware  and @tjscience  can review. I suggest testing this compared to other method to see which causes more system load (more threads created and or CPU usage).

Note: this relies on #146, and requires that #146 be merged in before this.


PCAssistSoftware:
For info

Can't build due to errors

'Statistic' does not contain a definition for 'NonZeroValue' and no accessible extension method 'NonZeroValue' accepting a first argument of type 'Statistic' could be found (are you missing a using directive or an assembly reference?)	

RoboQueueProgressEstimator.cs	lines 167	and 179

image

RFBomb:
Pull in the merged changes from dev. I only
Committed the one file to simplify the preview of changes files. (Hence the comment of requiring that merge first)

PCAssistSoftware:
> Pull in the merged changes from dev. I only Committed the one file to simplify the preview of changes files. (Hence the comment of requiring that merge first)

Okay, sorry I misunderstood....

RFBomb:
No problem, just trying to avoid the cluster of confusion that the RoboQueue pr 'changed files' became

PCAssistSoftware:
Thats understandable. How can I merge in the changes from dev just on my local machine using Github Desktop to test it? not something I have done before

RFBomb:
I typically open up the repo using Visual Studio, which you have already
done, then use GithubDesktop for the granular line-by-line commits and
ability to merge/pull since it’s better than VS for those things.
(Complicated merges I actually have GithubDesktop open the file in Visual
Studio Code to perform the merge outside of VS)

But to do it, open it up in GitHub Desktop, then in one of the drop downs
on top you should see ‘update from dev’, which should pull in everything
from official dev, not anyones branches/forks.

PCAssistSoftware:
Thanks, will take a proper look at it tomorrow then

PCAssistSoftware:
Have been doing some testing today with both this method and current method in Dev.

The method in Dev shows RoboSharp using around 52mb of Memory and peaks at about 37% CPU load.

The method in this PR shows RoboSharp using around 49mb of Memory and peaks at about 31% CPU load.

Now I am running on a pretty powerful machine (Intel Core i7 / 32gb RAM) so may notice even more difference if tested on a slower machine, but definitely shows less CPU usage consistently for the method in this PR.

I have repeated test multiple times using both versions and using different sizes of source files.

RFBomb:
Now I'm curious what the loads were prior to ProgressEstimator, or simply with it disabled / no UI updates. 

But that's kind-of what I anticipated occurring. And obviously this will change with usage / how much work robocopy is doing and how much files are in the log/how quick files are added to the log. 

But considering that ProgressEstimator only pushes updates out every 60ms or so at the quickest, it shouldn't be that much worse. 

I just thought of something though, let me do 1 commit and we can test again. This may be even better.

RFBomb:
If that change appears to reliably improve resource usage (lower memory/CPU utilization), I'll rig up ProgressEstimator the same way for testing

edit: On second thought, I don't think PE should be hooked up like this, because if too many Commands are running, then you potentially have UI updates, RoboQueueProgressEstimator updates (which can then trigger other UI updates or other events), chaining off the Process.OutputDataReceieved event, making that take way longer than needed, and we get back to the issue I was seeing where process exits but the list isn't filled in with entire log.

So I think that, while it may be worth testing, the current method (having a 'worker task' for updating the public objects) is likely the best way for ProgressEstimator to function. But this should be fine for RQPE because only those UI-Update worker tasks will be potentially waiting for the sync locks to occur, not the processing of the log lines.

PCAssistSoftware:
Okay when I get back home in a couple of hours I will repeat same tests as before and report back

PCAssistSoftware:
Latest PR doesn't build? 

Error	CS0103	The name 'Cancel' does not exist in the current context	C:\Users\darre\Documents\GitHub\RoboSharp\RoboSharp\Results\RoboQueueProgressEstimator.cs	228	


image

PCAssistSoftware:
Assumed `Cancel` should be `CancelTasks` on line 228? it seemed to build and work fine - so gone with that for testing.

Same jobs as yesterday with this PR gives me peak usage of 50mb RAM 28% CPU - ran three times and averaged results.

RFBomb:
Yea, that was using VSCode to edit the file in a vacuum, so I had no intellisense to catch the issue. You assumed correct. 

So that's good, memory within margin of error, but I was aiming for lower cpu overhead, so compared to dev, that's a huge improvement.

PCAssistSoftware:
> Yea, that was using VSCode to edit the file in a vacuum, so I had no intellisense to catch the issue. You assumed correct.

 Code fix pushed

> So that's good, memory within margin of error, but I was aiming for lower cpu overhead, so compared to dev, that's a huge improvement.

Yes definitely and improvement, and the source I was using was quite large - 85.6Gb / 85000 files / 14000 dirs

Tested both in BackupApp and in my own GUI - results comparable both ways

RFBomb:
Before this gets merged, I'd like to try one more thing and have you retest it. Wait for my next comment before the merge

PCAssistSoftware:
Yes that's fine - am free today so can test whatever is needed

RFBomb:
My plan is to tweak the ProgressEstimator so it's hopefully just a bit more efficient, doing the update time check inline with the math, instead of constantly waking and sleeping it, only wake if there's something to report.

PCAssistSoftware:
> My plan is to tweak the ProgressEstimator so it's hopefully just a bit more efficient, doing the update time check inline with the math, instead of constantly waking and sleeping it, only wake if there's something to report.

That make sense

PCAssistSoftware:
> Now I'm curious what the loads were prior to ProgressEstimator, or simply with it disabled / no UI updates.

Me too.  With it just disabled e.g. AddHandler line commented out - I noticed hardly any difference at all, but is this because it is still "running" in background even though not updating UI?

RFBomb:
I think the only way to test that would be going back before it was introduced. I wouldn't worry about it.

PCAssistSoftware:
Just been playing around with my GUI and am confused by the content of `e.CurrentDirectory.Name` in the `OnCopyProgressChanged` event.

Just added in a console.writeline to output all values returned to the console and was expecting to see it change as it goes through all of the SOURCE directory paths, but some of the source paths are never shown (despite files copying okay), and for some reason it then repeatedly lists just one of the DESTINATION paths.

I have two jobs setup:-

JOB1
Source = C:\Users\darre\Desktop\1a
Destination = C:\Users\darre\Desktop\2

image

N.B. each of the four subfolders here just contain 1 file each.

JOB2
Source = C:\Users\darre\Desktop\1b
Destination = C:\Users\darre\Desktop\2

image

So I would expect e.CurrentDirectory.Name` in the `OnCopyProgressChanged` event to **roughly** show something like below as it moves through each of the source directories:-

```
C:\Users\darre\Desktop\1a
C:\Users\darre\Desktop\1a\$foldername
C:\Users\darre\Desktop\1a\._folder
C:\Users\darre\Desktop\1a\.folder
C:\Users\darre\Desktop\1a\dazza's stuff
C:\Users\darre\Desktop\1b
```

But actually output is as below - trimmed to remove lots of lines which just contain repeat of previous path (number in brackets shows how many times that line appeared)

```
C:\Users\darre\Desktop\1a\					(2560)
C:\Users\darre\Desktop\1a\$foldername\		(6)
C:\Users\darre\Desktop\1a\._folder\
C:\Users\darre\Desktop\1a\dazza's stuff\
C:\Users\darre\Desktop\2\dazza's stuff\		(2568)
```

So as you can see it moved through first jobs directories okay but missed one of them (C:\Users\darre\Desktop\1a\.folder) although files copied okay.

Then which is where I am really confused it then jumped to repeatedly showing a **destination** sub-directory (C:\Users\darre\Desktop\2\dazza's stuff\) and never showed the source directory for JOB2 (C:\Users\darre\Desktop\1b)

N.B. I have no real need to show this information in my form, but was just checking each and every value returned to learn more and to ensure all working as expected, so only reporting it in case it is an issue? and to understand why discrepancy.

PCAssistSoftware:
> I think the only way to test that would be going back before it was introduced. I wouldn't worry about it.

okay

RFBomb:
The event is simply pulling in whatever the last processed Dir and File objects were that robocopy dropped into the log, which were then sent over to ProgressEstimator. 

For larger files (say one that is large enough to report 5 or 6 progress updates < 100%), those should all report the same dir and file in the event. Its definitely not something you should be writing to a log based on that, since you would get duplicates as youve seen. It would be more for updating a UI label to display file/dir strings.

PCAssistSoftware:
> The event is simply pulling in whatever the last processed Dir and File objects were that robocopy dropped into the log, which were then sent over to ProgressEstimator.
> 
> For larger files (say one that is large enough to report 5 or 6 progress updates < 100%), those should all report the same dir and file in the event. Its definitely not something you should be writing to a log based on that, since you would get duplicates as youve seen. It would be more for updating a UI label to display file/dir strings.

Yes I was only using it to show in GUI and wouldn't use it for logging etc. 

Was just curious as to why it never shows two of the source directories at all, especially when one of these source paths (C:\Users\darre\Desktop\1b) contains 5 x 512mb files , and why then it repeats destination directory thousands of times.

But if not a problem then don't worry, will avoid using it in my GUI so as to not confuse - and will make myself a note.

RFBomb:
Honestly, not sure what thats about. Based on the test I just did, which you can duplicate here, it seems to be working properly. 
If you want to compare logs for CopyProgressChanged vs OnFileProcessed, use the attached file (Don't commit it, because this shouldn't  be part of the official app!)

I updated the ProgressIndictaor to log the dirs and files inst List<String> objects. Modify Line 187 to some folder path you want to write to. Then you can compare these logs to see whats its doing and in what order. 

Note that this should only be running in SINGLE job mode, since this change affects all of those controls, and running multijob will likely cause an issue when two finished jobs trigger writing to the same files.

[MultiJob_CommandProgressIndicator.xaml.zip](https://github.com/tjscience/RoboSharp/files/7996277/MultiJob_CommandProgressIndicator.xaml.zip)


Edit: This doesn't clear the lists when a job is started, creating really long logs. 

Insert this on line 69: `SetupListObjs();`

Add this method on line 76
```
private void SetupListObjs()
        {
            Dirs = new List<string> { "Dirs Reported by OnFileProcessed", "---------------------" };
            Files = new List<string> { "Files Reported by OnFileProcessed", "---------------------" };
            Dirs2 = new List<string> { "Unique Dirs Reported by CopyProgressChanged", "---------------------" };
            Files2 = new List<string> { "Unique Files Reported by CopyProgressChanged", "---------------------" };
        }
```

For me, its outputting identical logs, so it appears to be working properly.

RFBomb:
Also, the last commit I did was the change I was hoping you could test out to somewhat optimize ProgressEstimator's whole UpdateTask. I don't expect any significant improvent  on that, but it might be somethingl

PCAssistSoftware:
Okay, very odd as can replicate this issue time and time again on various test sources.

More info below which may help:-

First issue where it was not listing a directory in Job 1 - caused because folder just contained a 0 byte file - can prove that is cause as if I remove 0 byte file then it works fine, if I create more folders containing 0 byte files then they too are not listed.

Second issue happens every single time - attached is a really small source/destination which I can replicate issue every time with, so add **TWO** jobs - first copying from 'test1' to '2' and second copying from 'test2' to '2' and make sure /e include sub directories is set.  Perhaps try this yourself and see what happens.

Is attachment correct in last response as file seems same as I already have? 

[test_files.zip](https://github.com/tjscience/RoboSharp/files/7996657/test_files.zip)

PCAssistSoftware:
> Also, the last commit I did was the change I was hoping you could test out to somewhat optimize ProgressEstimator's whole UpdateTask. I don't expect any significant improvent on that, but it might be somethingl

After last commit using same sources/destinations as before - 58mb RAM 31% CPU - so slightly higher figures.

I did note that this time if I repeat test several times (without closing app inbetween) that RAM increases, so first run was 58mb, then second was 79.5mb and third was 93.6mb

RFBomb:
Ok then it seem ProgressEstimator commit should be reverted and go back to the other way. Worth a shot anyway. 

0kb size files are a known issue, and I don't think anything can be done about it because robocopy simply creates the file without triggering progress change event. So that's expected, as far as I'm concerned. 

My test folder was the 'RoboSharp' folder from the solution. I had copied it to my desktop for the source, and went to a different folder for the destination. Copy was performed in about 6 seconds each time, logs identical each run.

If it's an issue, it's not happening on my machine at all.

PCAssistSoftware:
> Ok then it seem ProgressEstimator commit should be reverted and go back to the other way. Worth a shot anyway.

Okay, need to retest last version to see if RAM increasing happens then - but think it through I suppose that it as expected as app storing data from previous runs, so usage will increase each time I run it.

> 0kb size files are a known issue, and I don't think anything can be done about it because robocopy simply creates the file without triggering progress change event. So that's expected, as far as I'm concerned.

That's fine - at least we now know cause for issue 1 I had.

> My test folder was the 'RoboSharp' folder from the solution. I had copied it to my desktop for the source, and went to a different folder for the destination. Copy was performed in about 6 seconds each time, logs identical each run.
> 
> If it's an issue, it's not happening on my machine at all.

Did you try with my test files I supplied? as with those I can replicate issue on 3 different computers here

PCAssistSoftware:
N.B. my second issue ONLY happens if running more than ONE job, so please try the test files I provided and add the TWO jobs and you should see same issue.  If I just run each of the sources separately then issue doesn't occur, it is only when both jobs added and run together.

RFBomb:
Run together as in at same time? ConcurrentJobs = 2?

PCAssistSoftware:
> Run together as in at same time? ConcurrentJobs = 2?

Sorry, badly worded, I just meant in same queue - ConcurrentJobs is set to default of 1

RFBomb:
I've modified the logging function to be easier to read, and also introduced a different log to output the files and dirs in the order they occur. Here is the updates cs file.
[MultiJob_CommandProgressIndicator.xaml.zip](https://github.com/tjscience/RoboSharp/files/7999090/MultiJob_CommandProgressIndicator.xaml.zip)


Attached are the logs from OnFileProcessed and OnCopyProgressChange. I'm not seeing an issue.
Job1 runs, and you can see all folders created. 
Job 2 runs, but since all files are in root, OnCopyProgressChanged only reported the one directory. 

[Logs.zip](https://github.com/tjscience/RoboSharp/files/7999092/Logs.zip)

Its reporting the last known folder to be reported by RoboCopy. Its just that it happens to be an EXTRA folder instead of the SOURCE folder. I think thats your complaint. If thats the case, feel free to submit the change. if its exclusive to EXTRAS, shouldn't be a hard fix. Extras alraedy don't contribute to the 'Total', so this wouldn't be out of the norm here as far as logic is concerned

PCAssistSoftware:
Hi Robert

Will take a better look in the morning as getting late here and I am tired, but I think you have hit the nail on the head to be honest - I had just added the below line to my RoboCommand_OnCopyProgressChanged event so was only seeing the directories.

`Console.WriteLine("** " & e.CurrentDirectory.Name)`

After seeing your mention of it being an EXTRA folder it all makes sense now and I was misinterpreting what I was seeing from `e.CurrentDirectory.Name`

Just changed my line to below and it has just clicked!  I should have thought to look at `e.CurrentDirectory.FileClass` in the first place rather than just `.Name`!

Console.WriteLine("** " & e.CurrentDirectory.FileClass & " - " & e.CurrentDirectory.Name)

I will try your amended cs file tomorrow just to clarify the point but I think your summary does indeed explain what I was seeing.

I am really sorry to have confused matters - I just hadn't thought it through fully.

The logging you have added for testing this may well be a handy feature to have internally for debugging / testing purposes anyway, and I appreciate you taking the time to bear with me on this.

Will report back tomorrow :)

PCAssistSoftware:
Just tested again with your logging function and as per last nights reply you are correct, it was EXTRA folders I was seeing which confused me, which I would have known if I had thought to check `e.CurrentDirectory.FileClass` rather than just `e.CurrentDirectory.Name`! - so **sorry** for that.

Think the logging code you added to help though is very handy for debugging / testing purposes so perhaps should make it to Dev - even if just commented out so we can enable it as and when needed when testing changes etc?  thoughts?

>  If thats the case, feel free to submit the change. if its exclusive to EXTRAS, shouldn't be a hard fix. Extras already don't contribute to the 'Total', so this wouldn't be out of the norm here as far as logic is concerned

Don't think anything needs to change, as it was my confusion rather than an issue - as all becomes much clearer once you look at `.FileClass` rather than just `.Name`.

Thanks again

Let me know if you want anything else tested with this PR once you revert any changes you wanted etc based on last test results.  Happy to do any testing again today or over weekend.

RFBomb:
I wrapped the logging stuff into an IF statement, but its part of this commit now. So If you wanted to use it, just set the statement on line 224 to TRUE.

I also put a bool check in ProgressEstimator.AddDir to avoid the confusion. (I had thought about it, and yea, it was an easy fix to avoid a lot of confusion, so I did it. this comment took longer than the fix did)

image


And doing Test2 THEN test1:

image

PCAssistSoftware:
Excellent - that is perfect thanks, and is definitely proving useful during testing.

N.B. You accidently included 'MultiJob_CommandProgressIndicator.xaml.zip' in the last push

Is this ready to merge or did you still need to revert some previous changes as per https://github.com/tjscience/RoboSharp/pull/147#issuecomment-1029215554

RFBomb:
I'm testing now, would you mind doing the test again with current changes before we decide to revert? 
With VS running, I'm getting 100mb ram usage with the BackupApp and my largest spike was 25% CPU usage with a job size of 40.3MB, 1290 Files, 41 folders.

RFBomb:
I set my test up for what my application will be performing as a typical use case. I'm doing my testing with SINGLE JOB to test purely the ProgressEstimator change. 

Based on my testing, and using a simple await `Task.Delay(UpdatePeriod)`, all methods are negligible/ margin of error for CPU usage. That said, I think the current method in this PR provided the lowest CPU usage consistently? Not really sure, the CPU usage graph for all tests was hovering around 25% during most of my tests, but all methods were spiking between 38% and 42% ( 42% spikes only seen on method in dev!, while 33-38% and lower were seen in this PR. `Await Task.Delay` comparable spikes, while being overall less complex as far as code was concerned.

PCAssistSoftware:
Now doing some detailed testing with both my GUI and BackupApp both with this PR and with DEV - will be back very shortly with some results....

RFBomb:
The primary advantage I can see over the more complex method in this PR vs `await Task.Delay()` would be the task has no reason to run while large files are copying. At 7-8 updates per second, if a file takes 10 seconds to copy, thats 70-80 times the loop didn't need to run. So many small files it has to check the dateTime evaluation more often, but as files take longer to copy, the 'AddDir' and 'AddFile' events occur less often, meaning the update task runs less often too (as needed), instead of on a guaranteed interval.

Considering that, I think the overall benefit is worth the small bit of complexity, since it doesn't seem to have any significant performance impact for many quick items.

PCAssistSoftware:
No problem at all

Testing with same source as before which is 85.6Gb / 85000 files (approx) / 14000 dirs (approx)

Tested this time with both BackupApp and My test GUI.

All tests run in Visual Studio rather than running app standalone.

Tested using multi-job with two jobs added.

### **BackupApp - current PR**

Sitting idle before starting job

image

When running job it peaked at CPU 26.4% and MEMORY 82.9MB (increase of 48MB from Idle)

Closing and running same job again

CPU peaked at 27.3% and MEMORY 85.2MB (increase of 50.3MB from Idle)

N.B. I wondered why MEMORY usage was higher in today's testing than it was in previous testing and thought perhaps it is because of the storing of activity in the new SetupListObjs variables - so I commented out lines 71, 166-183 and 195-204 and ran again a couple of times - averages now as below

CPU peaked at 27.7% and MEMORY 65.8Mb (increase of 30.9MB from Idle)

### **BackupApp - DEV**

Sitting idle before starting job

image

When running job it peaked at CPU 30.5% and MEMORY 74.4MB (increase of 40.2MB from Idle)

Closing and running same job again

CPU peaked at 34.3% and MEMORY 66.5MB (increase of 32.3MB from Idle)

### **My test GUI using DLL from PR**

Sitting idle before starting job

image

When running job it peaked at CPU 32.7% and MEMORY 47.6MB (increase of 23.6MB from Idle)

Closing and running same job again

CPU peaked at 32.6% and MEMORY 46.8MB (increase of 22.8MB from Idle)

### **My test GUI using DLL from DEV**

Sitting idle before starting job

image

When running job it peaked at CPU 34.5% and MEMORY 56.5MB (increase of 32.5MB from Idle)

Closing and running same job again

CPU peaked at 35.3% and MEMORY 60.3MB (increase of 36.3MB from Idle)

----

### **Summary**

1) PR is definitely better than DEV for both CPU and MEMORY usage / peak

2) The new code added for "debug logs" does seem to increase MEMORY usage as even if we are not writing the logs on line 224 onwards we are still filling the variables on lines 166-183 and 195-204, so I propose that we set a boolean called something debugMode and we wrap lines 71, 166-183, 195-204 in `if (debugMode == true)` and change 224 to use this new bool - see changes I have pushed which I have tested and work for me
 
----

### **Response to comments**

> The primary advantage I can see over the more complex method in this PR vs await Task.Delay() would be the task has no reason to run while large files are copying. At 7-8 updates per second, if a file takes 10 seconds to copy, thats 70-80 times the loop didn't need to run. So many small files it has to check the dateTime evaluation more often, but as files take longer to copy, the 'AddDir' and 'AddFile' events occur less often, meaning the update task runs less often too (as needed), instead of on a guaranteed interval.
Considering that, I think the overall benefit is worth the small bit of complexity, since it doesn't seem to have any significant performance impact for many quick items.

I completely agree, I am often using it to copy a lot of larger files so will definitely better like this

RFBomb:
> PR is definitely better than DEV for both CPU and MEMORY usage / peak

Good, that's what I was aiming for. 

> The new code added for "debug logs" does seem to increase MEMORY usage as even if we are not writing the logs on line 224 onwards....

Yea, I wasn't really concerned with that too much, since (afaik) BackupApp is primarily for testing the library, and not meant as a real standalone application to on, so I wasnt really concerned with a memory increase there. (That logging stuff only affects the backup app, not the library itself)

That said, wrapping it into a higher level Boolean statement won't hurt anything, so if you did it and it works, then that's that.

PCAssistSoftware:
> > PR is definitely better than DEV for both CPU and MEMORY usage / peak
> 
> Good, that's what I was aiming for.
> 
All good then - let me know if you think we can merge it now?

> > The new code added for "debug logs" does seem to increase MEMORY usage as even if we are not writing the logs on line 224 onwards....
> 
> Yea, I wasn't really concerned with that too much, since (afaik) BackupApp is primarily for testing the library, and not meant as a real standalone application to on, so I wasnt really concerned with a memory increase there. (That logging stuff only affects the backup app, not the library itself)
> 
> That said, wrapping it into a higher level Boolean statement won't hurt anything, so if you did it and it works, then that's that.

Yes aware only affected app not library but just thought it made it tidier and covering all bases really :)

RFBomb:
If you already did that bool change you suggested,then I'm good with the merge.

PCAssistSoftware:
> If you already did that bool change you suggested,then I'm good with the merge.

Yes already done - will merge now

RFBomb:
Ok, so now, and I know I've said this before, I think that's all outstanding items regarding RoboQueue and RoboCommand and ProgressEstimator that we know of, and it's also been tested in you app and my app now. 

ProgressEstimator introduced a lot of kinks (it sucks we can't access the internal counters robocopy process itself uses, because we basically did the same thing as it's doing), but it is what it is. 

So now I'm looking forward to the next nuget release, so I can put the nuget reference back in my app, instead of a local dll path (not that that matters while the team that was supposed to help me with it is currently doing Jack-all due to being a man down)

PCAssistSoftware:
Yes agreed, I have some more work to do on my app over the next few days so will do one final run through of everything just to make sure - but pretty much think we (mainly you) have covered everything now.

Might have to wait for @tjscience for the nuget release as not sure that is something I have permissions to do - and only fair that he checks it over and is happy first anyway.

Thanks again for all your input on this, the changes and improvements have made something already very good into something excellent.

I can now get on and update all my existing apps that use it and start on some new ones I have planned.

Have a good weekend.

PCAssistSoftware:
@RFBomb - quick unrelated question - I know we have StartTime, EndTime and TimeSpan but is this per individual job (RoboCommand) or for all jobs (RoboQueue), or both?  Just asking before I dissect code to find out myself in case you knew off hand.  If for both then where do I obtain it, as want to be able in my results grid to show per job and overall if possible. Thanks

RFBomb:
Depends which event your looking at. 

OnCommandCompleted - RoboCommand

RunCompleted - RoboQueue

PCAssistSoftware:
> Depends which event your looking at.
> 
> OnCommandCompleted - RoboCommand
> 
> RunCompleted - RoboQueue

perfect - got it now - thanks :)

PCAssistSoftware:
Okay so got it showing for each job as it is available in OnCommandComplete like you said (e.StartTime, e.EndTime and e.TimeSpan.TosString)

But the rest of my grid information is added after queue completes so I obtain it from my ListResults or RunResults e.g.

```
Dim RunResults As IRoboCopyResultsList = Await queue.StartAll()
msgbox(RunResults.FilesStatistic.Total)
```

And I can't see StartTime, EndTime or TimeSpan in here?

Is there away to obtain them at this point, or can I only get them in RunCompleted

RFBomb:
That's actually a good point. 

I can't recall if it's already done, but maybe the DateTime for start/Stop should be added to the RoboCopyResults object itself if it's not in there. 

They are already recorded for the event args, so they should just be passed to the results builder and applied to the results object. Should be easy change to do. 

As far as the RoboQueue is concerned, currently RunCompleted is the only place to get that information. I don't think it makes sense to have it as part of the RoboCopyResultsList object, since that can be utilized by consumers. 

But maybe a new object that is derived from the resultsList that is exclusively created by RoboQueue could provide all that.

PCAssistSoftware:
No it isn't currently available anywhere in the results I see after running

`Dim RunResults As IRoboCopyResultsList = Await queue.StartAll()`

And this is where for my application I am obtaining all my "overall" results to show (for testing purposes initially) in a grid on my form - so the **only** thing missing from RunResults that I can see is overall StartTime, EndTime and TimeSpan.

So if they could be added I would appreciate it then I can get all information I need in my RunResults.

```
Dim recordToUpdate = resultsData.First(Function(p) p.JobName = "TOTALS")
        If recordToUpdate Is Nothing Then
            Return
        End If
        recordToUpdate.FilesTotal = RunResults.FilesStatistic.Total
        recordToUpdate.FilesSkipped = RunResults.FilesStatistic.Skipped
        recordToUpdate.FilesCopied = RunResults.FilesStatistic.Copied
        recordToUpdate.FilesMismatch = RunResults.FilesStatistic.Mismatch
        recordToUpdate.FilesFailed = RunResults.FilesStatistic.Failed
        recordToUpdate.FilesExtras = RunResults.FilesStatistic.Extras
        recordToUpdate.DirectoriesTotal = RunResults.DirectoriesStatistic.Total
        recordToUpdate.DirectoriesSkipped = RunResults.DirectoriesStatistic.Skipped
        recordToUpdate.DirectoriesCopied = RunResults.DirectoriesStatistic.Copied
        recordToUpdate.DirectoriesMismatch = RunResults.DirectoriesStatistic.Mismatch
        recordToUpdate.DirectoriesFailed = RunResults.DirectoriesStatistic.Failed
        recordToUpdate.DirectoriesExtras = RunResults.DirectoriesStatistic.Extras
        recordToUpdate.BytesTotal = Format_Size(RunResults.BytesStatistic.Total)
        recordToUpdate.BytesSkipped = Format_Size(RunResults.BytesStatistic.Skipped)
        recordToUpdate.BytesCopied = Format_Size(RunResults.BytesStatistic.Copied)
        recordToUpdate.BytesMismatch = Format_Size(RunResults.BytesStatistic.Mismatch)
        recordToUpdate.BytesFailed = Format_Size(RunResults.BytesStatistic.Failed)
        recordToUpdate.BytesExtras = Format_Size(RunResults.BytesStatistic.Extras)
        recordToUpdate.BytesPerSec = RunResults.SpeedStatistic.BytesPerSec
        recordToUpdate.MegaBytesPerMin = RunResults.SpeedStatistic.MegaBytesPerMin
```

image

RFBomb:
```
Public interface IRoboQueueResults : IRoboCopyResultsList
{
DateTime StartTime {get;}

DateTime StopTime {get;}

}
```

```
Public RoboQueueResults: RoboCopyResutsList, IRoboQueueResults
{
Internal RoboQueueResults():base(){}

Public DateTime StartTime {get; internal set;}

Public DateTime StopTime {get; internal set;}

}
```

Change task return type to that interface, change the object type to that object. Should work I think. 

Obviously these will have to be added to the interface / results folders, and the date time set prior to returning the task results.

PCAssistSoftware:
Sorry can't see where to make changes, when you get a spare moment - no hurry - perhaps you will take a look - this should hopefully be last piece of jigsaw for me.

N.B. It will need TimeSpan as well as StartTime / EndTime if possible so is same as what we get for each job in RoboCommand, but this time for RoboQueue

image

RFBomb:
Create new files for the new interface and class I spelt out in previous comment in the interface and results folder respectively. 

https://github.com/tjscience/RoboSharp/blob/1a00c8f0ac1bb8daf8a811e5e88835c9f70bf170/RoboSharp/RoboQueue.cs#L343

Type Should be new interface type. 
Line 344 should be RoboQueueResults type. 

Same for 341 and 352. 

https://github.com/tjscience/RoboSharp/blob/1a00c8f0ac1bb8daf8a811e5e88835c9f70bf170/RoboSharp/RoboQueue.cs#L507

507,519,534 should have return type Modified to be that interface. Same for 545,556, 570

https://github.com/tjscience/RoboSharp/blob/1a00c8f0ac1bb8daf8a811e5e88835c9f70bf170/RoboSharp/RoboQueue.cs#L554
Needs to be modified to apply StartTime to the ResultsObject. 
End time will need to be applied prior to invoking RunCompleted. (Same changes for StartAll_ListOnly)

https://github.com/tjscience/RoboSharp/blob/1a00c8f0ac1bb8daf8a811e5e88835c9f70bf170/RoboSharp/EventArgObjects/RoboQueueCompletedEventArgs.cs#L23
Needs to be changed to be return type of RoboQueueResults

RoboQueueResults also should have Clone() overridden to clone itself including the DateTime properties.

PCAssistSoftware:
Okay thank you - will give it a go tomorrow and let you know how I get on

PCAssistSoftware:
I have taken a look, but think this is beyond my abilities to be honest - remember I don't program in C# which makes it harder. 

Will try again tomorrow with fresh eyes, but if not I have it working for now simply setting three temporary variables to hold StartTime, EndTime and TimeSpan results in RunCompleted, so I can then access them later to populate overall results in my grid.

RFBomb:
I'll see if I have a chance to set it up this week

PCAssistSoftware:
On a separate note (will open issue later if needed)

I think changes / improvements may be needed to log parsing for estimator as just been reading some RoboCopy documentation and doing some tests with switches such as /IS (include same)

Log shows `same` (all in lower case) normally, but if using /IS then the S is capitalized (`Same`) - we only handle it correctly if lowercase (according to what I see in RoboSharpConfiguration.cs) - so this affects values for copied / skipped.

Uppercase S means copied (e.g. using /IS), lowercase s means skipped (default behaviour)

N.B. same logic possibly applies to other switches such as /XO (exclude older) and /XN (exclude newer) as we only handle Newer / Older and not newer / older.

Still looking at /IT (include tweaked), /XC (exclude changed) as these can return Tweaked / tweaked or Changed / changed but not sure how (or if) we handle these currently - but certainly dependent on switches and upper/lower case first letter would make difference to results being correctly counted as copied or skipped.

Really handy reference to all the tags here - https://theether.net/download/Microsoft/Utilities/robocopy.pdf - is for older version of RoboCopy but just tested and all still relevant to current version - **see page 30**.

Also shows lots of tags I never even knew about which again we don't handle but could affect results shown e.g. large, small, too new, too old, attrib when using switches such as /MIN, /MAX, /MINAGE, /MAXAGE, /XA

>Each line in the output log begins with a brief text tag, which is formatted according to the following rules:
> 
> · All capital letters indicate an anomaly that should be investigated.
> · Initial capital letters indicate a file that was selected for copying.
> · All lowercase letters indicate a file that was skipped (displayed only if the /V switch is used)

RFBomb:
This should only Affect ListOnly operations, as OnCopyProgressChanged lets PE know if a file has started/finished copying regardless of what it identified as for XO/XN/Same/etc. 

I've provided the base logic for it, if you or someone else wants to expand on that pattern to include all the possible tags (which I think is overkill unless we also include the robocopy language files to identify the tags), the. That's fine. But expanding past what I've currently provided for that is something I'm not interested in doing, since it will primarily affect only ListOnly operations. 

Plus, it may already be handled:

```

if (currentFile.FileClass.Equals(Config.LogParsing_NewerFile, StringComparison.CurrentCultureIgnoreCase))
                    {
                        if (command.SelectionOptions.ExcludeNewer)
                            PerformByteCalc(currentFile, WhereToAdd.Skipped);
                        else
                            PerformByteCalc(currentFile, WhereToAdd.Copied);

```

PCAssistSoftware:
From the tests I have done it doesn't just affect ListOnly - see figures below comparing with /IS and without /IS - results are fine, just progress estimator which is wrong

image

Also a more immediate issue when using /IS for some reason in the OnCopyProgressChanged event `e.CurrentFile.Name` is nothing despite file obviously being copied as shows CurrentFileProgress

image

RFBomb:
Ok then update PE AddFile method to make it work.

PCAssistSoftware:
Okay will look this week and see what I can do, may as well try and cover all bases, no point offering the switches in GUI if it then breaks parts the estimator.

Any comment on second issue in last post as that seems a more serious problem potentially?

RFBomb:
That's because in PE it's being set to null because the check for 'same' doesn't follow the same pattern as XN and XO. If you look at the file, the issue should be pretty easy to see and update to a similar pattern 

(Fixing one should fix both - same root cause)

PCAssistSoftware:
> That's because in PE it's being set to null because the check for 'same' doesn't follow the same pattern as XN and XO. If you look at the file, the issue should be pretty easy to see and update to a similar pattern

Okay will spend some time on it this week, may come back for some advice if I get stuck, but more than happy to have a go

> (Fixing one should fix both - same root cause)

That's what I thought - good

PCAssistSoftware:
Issue #148 opened for any further discussion - will start working on PR this week

PCAssistSoftware:
@RFBomb 

Hi Robert

Hoping you could solve a problem for me.

Busy implementing latest RoboSharp in my app and all going very well - all your changes make it so much easier to achieve all the things i have always wanted to do in relation to showing progress information and running multiple jobs in a queue.

**BUT** I have now hit one weird thing during testing which I am hoping you can replicate and help fix - hoping it is quite simple.

Happens during some mirror jobs and I have created a very simple 1 source 1 destination example which demonstrates the exact problem in the BackupApp, along with sample files / job so you can hopefully replicate it for yourself.

Source : \source\mirror2
Destination \destination\mirror2

Switches: \e \mir

1 more files in destination than source so should correctly find extras 

When you run (even just list only) the progress estimator for the job is correctly returning byte information BUT the overall progress estimator is not returning / showing anything at all - as you can see in screenshot below.

So in my own GUI the BytesStatistic_PropertyChanged event is not firing at all and I need this event to reliably fire as I use it to update progress information in a grid on my form.  It fires pefectly fine on all other occassions (and for files/directories) and my app is working fine 99% of the time, just for some reason when running a job such as explained above it doesn't fire for bytes and i can replicate it every single time, and i don't know why.

Any thoughts please?  So close to finishing what I want to achieve now and just hit this issue....

image

[test.zip](https://github.com/tjscience/RoboSharp/files/8104222/test.zip)

PCAssistSoftware:
UPDATE:

One of the files used in this test is 0 bytes (I know sorry!!! :) - I realise I keep having issues with 0 byte files BUT for my usage scenario I will be encountering a lot of 0 byte files as they are used as pointer files for a specific app a company uses)

In this particular issue, which you can replicate as above, the 0 byte file is picked up as same and skipped so not really relevant as such, but does seem to play a part in the problem as if I replace it with a different file which isn't 0 bytes then problem goes away.

As you can see from above screenshot the **Progress Estimate** for the job itself works fine and show the correct bytes details, it just doesn't seem to pass those bytes details on to the **Total Estimate** so the BytesStatistic_PropertyChanged event then never fires and I don't get the results I need.

So it seems in this example where there are just two files, one is zero bytes and already exists so is reported as **same** so is skipped, the second file only exists in destination so is detected correctly as ***EXTRA** - all shows okay in **Progress Estimate**, but bytes details don't appear in **Total Estimate** and BytesStatistic_PropertyChanged event is never fired.

If you repeat same test but replace the zero byte file with a 1 byte file then it all works fine, so what stops the BYTE information appearing in **Total Estimate** and the BytesStatistic_PropertyChanged event firing when it is zero bytes but allow it to work fine when it is larger than zero bytes??

RFBomb:
> so what stops the BYTE information appearing in Total Estimate and the BytesStatistic_PropertyChanged event firing when it is zero bytes but allow it to work fine when it is larger than zero bytes??

The event only fires when the value is changed. Since X+0=X, the value never changes, thus the event never fires. 


I see two options: 

- ensure that the ByteStat math is always performed first then the FileStat event can be relied on to update the display for both File and Byte stats. 

- when the UI starts the job, the UI should be initialized by displaying the Value (which will be zero), and since it's never increased by a bunch of 0byte files, it will still display proper value.

PCAssistSoftware:
But surely the event should be firing for the *EXTRA file (4.rtf) which is 7 bytes

EDIT: or does it not fire for files which are being deleted? even though that causes a change in bytes?

> Files and Dirs In Order Reported by OnFileProcessed
> ---------------------
> 
> Existing Dir(1) - C:\Users\darre\Desktop\mirror2\
> *EXTRA File(7) - 4.rtf
> same(0) - 3.txt

In this particular case the 0 byte file may be the cause of the problem, but is not what actually made me notice the issue so to speak - what made me notice it was that my on screen figures were always 7 bytes out, and that was because BytesStatistic_PropertyChanged hadn't triggered for the *EXTRA file (4.rtf) so I hadn't then minused the size in my grid, so all other counts were correct just not the bytes.

In your two options the first makes more sense to me, is this something easily doable?

----

Just to explain what I have done in my GUI so the above usage makes sense - I run a 'list only' run and from the results I populate a grid with values for FilesStatistic.Copied, FilesStatistic.Extras, DirectoriesStatistic.Total, DirectoriesStatistic.Extra, BytesStatistic.Copied and BytesStatistic.Extra and then during a 'copy' run I catch the ProgressEstimator events from FilesStatistic_PropertyChanged, DirectoriesStatistic_PropertyChanged and BytesStatistic_PropertyChanged and minus the values from my totals in the grid so I can show a live countdown for files/directories/bytes being copied and deleted and progress bars for each - works really well for my needs - but obviously do need the event to always fire

This video shows it working perfectly - https://www.screencast.com/t/GbRjFbXtzlzy

This video is result of above issue when it doesn't fire and thus left still showing the 7 bytes it didn't fire for - https://www.screencast.com/t/VxK4BA2hEKc

Issue #148, opened 2022-02-06
Progress Estimator logic needs fixing for handling of switches such as /IS


PCAssistSoftware:
Initially discussed here - https://github.com/tjscience/RoboSharp/pull/147#issuecomment-1030865157

Key points of discussion copied below - will start working on PR this week.

------------

I think changes / improvements may be needed to log parsing for estimator as just been reading some RoboCopy documentation and doing some tests with switches such as /IS (include same)

Log shows same (all in lower case) normally, but if using /IS then the S is capitalized (Same) - we only handle it correctly if lowercase (according to what I see in RoboSharpConfiguration.cs) - so this affects values for copied / skipped.

Uppercase S means copied (e.g. using /IS), lowercase s means skipped (default behaviour)

~~N.B. same logic possibly applies to other switches such as /XO (exclude older) and /XN (exclude newer) as we only handle Newer / Older and not newer / older.~~

Still looking at /IT (include tweaked), /XC (exclude changed) as these can return Tweaked / tweaked or Changed / changed but not sure how (or if) we handle these currently - but certainly dependent on switches and upper/lower case first letter would make difference to results being correctly counted as copied or skipped.

Really handy reference to all the tags here - https://theether.net/download/Microsoft/Utilities/robocopy.pdf - is for older version of RoboCopy but just tested and all still relevant to current version - see page 30.

Also shows lots of tags I never even knew about which again we don't handle but could affect results shown e.g. large, small, too new, too old, attrib when using switches such as /MIN, /MAX, /MINAGE, /MAXAGE, /XA

> Each line in the output log begins with a brief text tag, which is formatted according to the following rules:
> 
> · All capital letters indicate an anomaly that should be investigated.
> · Initial capital letters indicate a file that was selected for copying.
> · All lowercase letters indicate a file that was skipped (displayed only if the /V switch is used)

PCAssistSoftware:
**Response from @RFBomb** 

This should only Affect ListOnly operations, as OnCopyProgressChanged lets PE know if a file has started/finished copying regardless of what it identified as for XO/XN/Same/etc.

I've provided the base logic for it, if you or someone else wants to expand on that pattern to include all the possible tags (which I think is overkill unless we also include the robocopy language files to identify the tags), the. That's fine. But expanding past what I've currently provided for that is something I'm not interested in doing, since it will primarily affect only ListOnly operations.

Plus, it may already be handled:

```
[...]

PR #149, opened 2022-02-06
Fixes for #148

PCAssistSoftware:
Testing of fixes for Progress Estimator logic needs fixing for handling of switches such as /IS #148


RFBomb:
I don't have commit permission here, but here is the change for IncludeSame
image


[ProgressEstimator.zip](https://github.com/tjscience/RoboSharp/files/8017162/ProgressEstimator.zip)

PR #150, opened 2022-02-07
Add TimeSpan to the Results Object and introduce RoboQueueResults

RFBomb:
RoboQueueResults implements IRoboCopyResultsList, so previous functionality still exists. But now has a TimeSpan for start/end time.

Note: EndTime is not accurate until after command completes. Same for TimeSpan.


PCAssistSoftware:
Just wanted to quickly say I had noticed this and really appreciate it, will test it later, just got bogged down with the other PR

PCAssistSoftware:
What am I missing here?

If I set a breakpoint after running

`Dim RunResults As IRoboCopyResultsList = Await queue.StartAll()`

I can see StartTime, EndTime, TimeSpan in properties of RunResults, but trying to access `RunResults.StartTime` is giving me an error saying it is not a member, even though I can clearly see it when it hits breakpoint

image

RFBomb:
`IRoboCopyResultsList ` should be `IRoboQueuResults` 

I had to create a new object to return because a consumer can use the RoboCopyResultsList on their own, without RoboQueue. Since RoboQueue has a known start/end time, while a generic list does not, a new object was created that houses and protects the results list RoboQueue generates.

PCAssistSoftware:
> `IRoboCopyResultsList ` should be `IRoboQueuResults`

Excellent - thank you - that works perfectly for me.

Just one thing (not that it really matters) but shouldn't IRoboQueuResults be IRoboQueu**e**Results e.g. with the e at end of Queue

P.S. Needs amending on wiki - https://github.com/tjscience/RoboSharp/wiki/RoboQueue

RFBomb:
Yea, wiki will have to wait, I have some other priorities at the moment.
The wiki will have to be gone through thoroughly though, object by object. If I recall, I think I had already done a decent chunk of the easier one like Statistic, but I don't think its 'done'

PCAssistSoftware:
> Yea, wiki will have to wait, I have some other priorities at the moment. The wiki will have to be gone through thoroughly though, object by object. If I recall, I think I had already done a decent chunk of the easier one like Statistic, but I don't think its 'done'

Yes thats fair enough, it is quite a time consuming job to do and to fit round the day job - I actually just meant initially changing `IRoboCopyResultsList` to `IRoboQueueResults` in the sample code snippet (which I have now done), I hadn't thought through all the other related changes that need doing - as you say that's a job for later.....

Thanks again for doing this, will run a couple more tests and then if you are are happy will merge it in

RFBomb:
As long as it works for you, and both RoboCopyResults objects and RoboQueueResults objects are showing the timestamps just fine, I say merge

PCAssistSoftware:
Yes all showing fine for me for both objects

image

Will merge now

PCAssistSoftware:
@RFBomb 

Not sure what has happened here but for some reason BackupApp from Dev won't run - confused as it was certainly running okay when testing this PR so must be since the last merge? although convinced I had run it multiple times before and after the merge, but just downloaded clean copy and hit below issue. Thoughts please?

Builds okay, but when running get error as below on line 76 of JobHistoryExpander_SingleJob.xaml.cs

image

Edit: in fact I know it was definitely running okay because I had commented on your GUI changes re the debugging check box etc..... so confused

RFBomb:
I have no idea. 
It's consistent?

When it occurs, is the `list` a RoboCopyResultList or is it a RoboQueueResult object that is being bound to? 

Either way though, both are a RoboCopyResultList (RoboQueueResult just protects it from outside influence).

My guess is that an error occurred during construction of the object, so it didn't finish initializing. But yea, it was working before the merge so I don't know. 

Edit:
Actually, I think I may know what's happening. `list` is itself null. A null object was passed into the method somehow. Set a conditional breakpoint to see when they submit a null value, then move the cursor to the end of that method and step out to see where the null Value is coming from.

PCAssistSoftware:
> It's consistent?

Yes, can't even get GUI to run

> When it occurs, is the `list` a RoboCopyResultList or is it a RoboQueueResult object that is being bound to?

Both list and Resultslist are null at the point the exception is thrown

IRoboCopyResultsList

> Edit: Actually, I think I may know what's happening. `list` is itself null. A null object was passed into the method somehow. Set a conditional breakpoint to see when they submit a null value, then move the cursor to the end of that method and step out to see where the null Value is coming from.

Tracing it back from point that the exception is thrown leads me to the below in the main window initialization code, but not sure how it helps or where to go from here, just confused as to how the PR worked perfectly fine before we (I) merged it

```
//Setup SingleJob Tab
SingleJobExpander_JobHistory.BindToList(SingleJobResults); 
```

RFBomb:
Is it being bound to before RoboQueue started? If so, then that may be the issue. 

- Previously, I had a single list RoboQueue was populating and clearing between started. Once start/end time was introduced, that really wasnt feasible, but also I realized it didn't make a whole lot of sense to have a results object available when nothing had been started. So now it creates the object when it starts running the queue, since that also initializes the 'StartTime' property. 

Move the bind methods to after it starts in the start button? (Use the checkbox to determine which box to bind to, list only or not). 

No idea why it was working when we tested it and broke now, but I think that's the issue.

RFBomb:
Secondary thought, maybe we can add the result object to the progress estimator created event since by the time the progress estimators created the new result object should be available as well

PCAssistSoftware:
> Is it being bound to before RoboQueue started? If so, then that may be the issue.

I assume it is being bound to at the point it is initializing the form as it is not even getting to the point of even displaying the BackupApp main window at all

The earliest point I can trace it back to is line 56 of MainWindow.xaml.cs

https://github.com/tjscience/RoboSharp/blob/825d871a10f3df671883a1cab1fd8d12c6a24611/RoboSharp.BackupApp/MainWindow.xaml.cs#L56

This then calls 

https://github.com/tjscience/RoboSharp/blob/825d871a10f3df671883a1cab1fd8d12c6a24611/RoboSharp.BackupApp/JobHistoryExpander.xaml.cs#L41-L54

And at that point resultsList (RoboCopyResultsList) is null

> * Previously, I had a single list RoboQueue was populating and clearing between started. Once start/end time was introduced, that really wasnt feasible, but also I realized it didn't make a whole lot of sense to have a results object available when nothing had been started. So now it creates the object when it starts running the queue, since that also initializes the 'StartTime' property.
> 
> Move the bind methods to after it starts in the start button? (Use the checkbox to determine which box to bind to, list only or not).
> 
> No idea why it was working when we tested it and broke now, but I think that's the issue.

When you get a spare moment could you please take a look or give me some clues as to what to change and where as struggling here i'm afraid

RFBomb:
Lines 64 and 67 I think are the culprit. 

Move them to line 428-ish.  If (checkbox) await Start* should also bind them. 

If {
Await starts;
Bind 
}
Else {
Await and Bind other }

PCAssistSoftware:
You are an absolute star - thank you

The below change does indeed allow the GUI to launch and the test jobs I have done worked as expected.

  ```
if (chkListOnly.IsChecked == true)
                { 
                    await RoboQueue.StartAll_ListOnly();
                    MultiJob_ListOnlyResults.BindToList(RoboQueue.ListResults);
                }
                else
                { 
                    await RoboQueue.StartAll();
                    MultiJob_RunResults.BindToList(RoboQueue.RunResults);
                }
```

Unless you can think of anything else I will commit this change now

RFBomb:
ThTs the exact change I was alluding to. Fine to commit by me.

PCAssistSoftware:
> ThTs the exact change I was alluding to. Fine to commit by me.

Thanks for your help :) - all done

Issue #151, opened 2022-02-15
MoveFilesAndFolders - "Cannot delete source Directory" error


RFBomb:
My project is taking ALL files and folders EXCEPT 2 folders from the Root of a USB stick, and stuffing them into a backup folder on the root of the USB. 

For example:

Source: E:\
Destination:  `$"E:\USB_BACKUP\{DateTime.Now.ToString("yyyy-MM-dd_hh_m_tt")};"`
```
E:\FolderIgnore1
E:\USB_BACKUP
E:\SomeOtherFolderWithFiles
E:\SomeFile

```
turns into 

```
E:\FolderIgnore1
E:\USB_BACKUP
E:\USB_BACKUP\CurrentDateTimeFolderName
E:\USB_BACKUP\CurrentDateTimeFolderName\SomeOtherFolderWithFiles
E:\USB_BACKUP\CurrentDateTimeFolderName\SomeFile
```

All that works, but RoboCopy reports an error that it cannot delete the source directory, which is typically done during the 'MoveFilesAndFolders'. I know WHY its erroring - It can't delete the source from the drive because the source IS the root of the drive.

Any idea how to prevent that error being generated?

RFBomb:
Also, it apparently moved the 'System Volume Information' folder o.0

PCAssistSoftware:
Can you upload a job file so I can see exactly what settings you are using, and a example directory structure so I can try and replicate it here?

As for "System Volume Information" then yes I would expect it to move that as well if the source is the root of a drive and you haven't added it as an excluded directory, also happens with $RECYCLE.BIN

RFBomb:
[BootVersion.zip](https://github.com/tjscience/RoboSharp/files/8075344/BootVersion.zip)
[JOB.zip](https://github.com/tjscience/RoboSharp/files/8075345/JOB.zip)

( I already added the SVI folder to ignored list)
Unzip 'Bootversion' onto the root of a USB drive. After backup is complete, it should exist within the 'USB_BACKUP' folder in a subfolder. 
Unzip BootVersion onto the root again, run the backup again and it goes into a new subfolder (Obviously you have to change the folder name, but my code does this automatically via DateTime.Now as the folder name)

PCAssistSoftware:
[...]

PR #152, opened 2022-02-21
IProgressEstimator.ValuesUpdated Event Added

RFBomb:
- Introduced the ValuesUpdated event to IProgressEstimator and implemented it.
- This change streamlines the UI Update process for ProgressEstimator, and also simplifies how RoboQueue gets and pushes out its own updates

@tjscience - See if this resolves your issues you noted earlier today/yestrday.

Update your bindings to use the new event provided by IProgressEstimator.

### DIFF
--- a/RoboSharp.BackupApp/MainWindow.xaml.cs
+++ b/RoboSharp.BackupApp/MainWindow.xaml.cs
@@ -502,16 +502,20 @@ namespace RoboSharp.BackupApp
 
         private void RoboQueue_OnProgressEstimatorCreated(RoboQueue sender, EventArgObjects.ProgressEstimatorCreatedEventArgs e)
         {
-            e.ResultsEstimate.DirectoriesStatistic.OnTotalChanged += DirectoriesStatistic_PropertyChanged;
-            e.ResultsEstimate.FilesStatistic.OnTotalChanged += FilesStatistic_PropertyChanged;
-            e.ResultsEstimate.BytesStatistic.OnTotalChanged += BytesStatistic_PropertyChanged;
+            e.ResultsEstimate.ValuesUpdated += IProgressEstimatorValuesUpdated;
         }
 
-        private void DirectoriesStatistic_PropertyChanged(Results.Statistic sender, System.ComponentModel.PropertyChangedEventArgs e) => UpdateLabel(ProgressEstimator_Directories, sender);
-        private void FilesStatistic_PropertyChanged(Results.Statistic sender, System.ComponentModel.PropertyChangedEventArgs e) => UpdateLabel(ProgressEstimator_Files, sender);
-        private void BytesStatistic_PropertyChanged(Results.Statistic sender, System.ComponentModel.PropertyChangedEventArgs e) => UpdateLabel(ProgressEstimator_Bytes, sender);
+        private void IProgressEstimatorValuesUpdated(Interfaces.IProgressEstimator sender, EventArgObjects.IProgressEstimatorUpdateEventArgs e)
+        {
+            Dispatcher.Invoke(() =>
+            {
+                UpdateLabel(ProgressEstimator_Directories, e.DirectoriesStatistic);
+                UpdateLabel(ProgressEstimator_Files, e.FilesStatistic);
+                UpdateLabel(ProgressEstimator_Bytes, e.BytesStatistic);
+            });
+        }
 
-        private void UpdateLabel(TextBlock lbl, RoboSharp.Results.Statistic stat)
+        private void UpdateLabel(TextBlock lbl, RoboSharp.Interfaces.IStatistic stat)
         {
             Dispatcher.Invoke(() =>
             {
--- a/RoboSharp.BackupApp/MultiJob_CommandProgressIndicator.xaml.cs
+++ b/RoboSharp.BackupApp/MultiJob_CommandProgressIndicator.xaml.cs
@@ -152,25 +152,17 @@ namespace RoboSharp.BackupApp
                 ProgressEstimator_Directories.Text = "Directories";
                 ProgressEstimator_Bytes.Text = "Bytes";
             });
-            e.BytesStatistic.PropertyChanged += ByteStats_PropertyChanged;
-            e.DirectoriesStatistic.PropertyChanged += DirStats_PropertyChanged;
-            e.FilesStatistic.PropertyChanged += FileStats_PropertyChanged;
+            e.ValuesUpdated += IProgressEstimatorValuesUpdated;
         }
 
-        private void FileStats_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        private void IProgressEstimatorValuesUpdated(Interfaces.IProgressEstimator sender, EventArgObjects.IProgressEstimatorUpdateEventArgs e)
         {
-            ProgressEstimator_Files.Dispatcher.Invoke(() => ProgressEstimator_Files.Text = ((RoboSharp.Results.Statistic)sender).ToString(true, true, "\n", true));
-        }
-
-        private void DirStats_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
-        {
-            ProgressEstimator_Directories.Dispatcher.Invoke(() => ProgressEstimator_Directories.Text = ((RoboSharp.Results.Statistic)sender).ToString(true, true, "\n", true));
-        }
-
-        private void ByteStats_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
-        {
-            ProgressEstimator_Bytes.Dispatcher.Invoke(() => ProgressEstimator_Bytes.Text = ((RoboSharp.Results.Statistic)sender).ToString(true, true, "\n", true));
-
+            Dispatcher.Invoke(() =>
+            {
+                ProgressEstimator_Files.Text = e.FilesStatistic.ToString(true, true, "\n", true);
+                ProgressEstimator_Bytes.Text = e.BytesStatistic.ToString(true, true, "\n", true);
+                ProgressEstimator_Directories.Text = e.DirectoriesStatistic.ToString(true, true, "\n", true);
+            });
         }
 
         #endregion
--- a/RoboSharp/EventArgObjects/IProgressEstimatorUpdateEventArgs.cs
+++ b/RoboSharp/EventArgObjects/IProgressEstimatorUpdateEventArgs.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using RoboSharp.Interfaces;
+using RoboSharp.Results;
+
+// Do Not change NameSpace here! -> Must be RoboSharp due to prior releases
+namespace RoboSharp.EventArgObjects
+{
+    /// <summary>
+    /// Event Args provided by IProgressEstimator objects to notify the UI it should refresh the stat values
+    /// </summary>
+    public class IProgressEstimatorUpdateEventArgs : EventArgs
+    {
+        /// <summary> Dummy Args with Values of 0 to perform final updates through ProgressEstimator without creating new args every time</summary>
+        internal static IProgressEstimatorUpdateEventArgs DummyArgs { get; } = new IProgressEstimatorUpdateEventArgs(null, null, null, null);
+        
+        private IProgressEstimatorUpdateEventArgs() : base() { }
+
+        internal IProgressEstimatorUpdateEventArgs(IProgressEstimator estimator, IStatistic ByteChange, IStatistic FileChange, IStatistic DirChange) : base()
+        {
+            Estimator = estimator;
+            ValueChange_Bytes = ByteChange ?? Statistic.Default_Bytes;
+            ValueChange_Files = FileChange ?? Statistic.Default_Files;
+            ValueChange_Directories = DirChange ?? Statistic.Default_Dirs;
+        }
+
+        /// <summary>
+        /// <inheritdoc cref="Results.ProgressEstimator"/>
+        /// </summary>
+        private IProgressEstimator Estimator { get; }
+
+        /// <inheritdoc cref="IProgressEstimator.BytesStatistic"/>
+        public IStatistic BytesStatistic => Estimator?.BytesStatistic;
+
+        /// <inheritdoc cref="IProgressEstimator.FilesStatistic"/>
+        public IStatistic FilesStatistic => Estimator?.FilesStatistic;
+
+        /// <inheritdoc cref="IProgressEstimator.DirectoriesStatistic"/>
+        public IStatistic DirectoriesStatistic => Estimator?.DirectoriesStatistic;
+
+        /// <summary>IStatistic Object that shows how much was added to the { <see cref="BytesStatistic"/> } object during this UI Update</summary>
+        public IStatistic ValueChange_Bytes { get; }
+
+        /// <summary>IStatistic Object that shows how much was added to the { <see cref="FilesStatistic"/> } object during this UI Update</summary>
+        public IStatistic ValueChange_Files { get; }
+
+        /// <summary>IStatistic Object that shows how much was added to the { <see cref="DirectoriesStatistic"/> } object during this UI Update</summary>
+        public IStatistic ValueChange_Directories { get; }
+
+    }
+}
+
--- a/RoboSharp/Interfaces/IProgressEstimator.cs
+++ b/RoboSharp/Interfaces/IProgressEstimator.cs
@@ -38,5 +38,8 @@ namespace RoboSharp.Interfaces
         /// </summary>
         RoboCopyExitCodes GetExitCode();
 
+        /// <summary> Event that occurs when this IProgressEstimatorObject's IStatistic values have been updated. </summary>
+        event ProgressEstimator.UIUpdateEventHandler ValuesUpdated;
+
     }
 }
--- a/RoboSharp/Results/ProgressEstimator.cs
+++ b/RoboSharp/Results/ProgressEstimator.cs
@@ -471,37 +471,33 @@ namespace RoboSharp.Results
             Statistic TD = null;
             Statistic TB = null;
             Statistic TF = null;
-            bool DirAdded;
-            bool FileAdded;
             lock (DirLock)
             {
-                DirAdded = tmpDir.NonZeroValue;
-                if (DirAdded)
+                if (tmpDir.NonZeroValue)
                 {
                     TD = tmpDir.Clone();
                     tmpDir.Reset();
                 }
-
             }
             lock (FileLock)
             {
-                FileAdded = tmpFile.NonZeroValue || tmpByte.NonZeroValue;
-                if (FileAdded)
+                if (tmpFile.NonZeroValue)
                 {
                     TF = tmpFile.Clone();
+                    tmpFile.Reset();
+                }
+                if (tmpFile.NonZeroValue)
+                {
                     TB = tmpByte.Clone();
                     tmpByte.Reset();
-                    tmpFile.Reset();
                 }
             }
             //Push UI update after locks are released, to avoid holding up the other thread for too long
-            if (DirAdded) DirStatField.AddStatistic(TD);
-            if (FileAdded)
-            {
-                ByteStatsField.AddStatistic(TB);
-                FileStatsField.AddStatistic(TF);
-            }
-            if (DirAdded | FileAdded)
+            if (TD != null) DirStatField.AddStatistic(TD);
+            if (TB != null) ByteStatsField.AddStatistic(TB);
+            if (TF != null) FileStatsField.AddStatistic(TF);
+            //Raise the event if any of the values have been updated
+            if (TF != null || TD != null || TB != null)
             {
                 ValuesUpdated?.Invoke(this, new IProgressEstimatorUpdateEventArgs(this, TB, TF, TD));
             }
--- a/RoboSharp/Results/ProgressEstimator.cs
+++ b/RoboSharp/Results/ProgressEstimator.cs
@@ -486,7 +486,7 @@ namespace RoboSharp.Results
                     TF = tmpFile.Clone();
                     tmpFile.Reset();
                 }
-                if (tmpFile.NonZeroValue)
+                if (tmpByte.NonZeroValue)
                 {
                     TB = tmpByte.Clone();
                     tmpByte.Reset();
--- a/RoboSharp/Results/ProgressEstimator.cs
+++ b/RoboSharp/Results/ProgressEstimator.cs
@@ -7,6 +7,7 @@ using RoboSharp.Interfaces;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Runtime.CompilerServices;
+using RoboSharp.EventArgObjects;
 
 namespace RoboSharp.Results
 {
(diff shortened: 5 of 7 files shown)


PCAssistSoftware:
Hi, thanks for this.

Certainly resolves the issue when running same test in BackupApp :)

Now trying to convert my GUI to use this new method but hitting a few issues as when I now run my app (even list only) I get error in Statistic.cs

image

Now trying to step through code to work out why..... will report back later as trying to fit debugging around paid work :)

PCAssistSoftware:
UPDATE:

The line which seems to cause the error is when I start the queue e.g.

`Dim ListResults As IRoboQueueResults = Await queue.StartAll_ListOnly()`

But only happens with ListOnly and for life of me I can't yet see why only happens with ListOnly

RFBomb:
can you post the call stack?

PCAssistSoftware:
Have managed to replicate the issue in a simple sample app for you which is attached

Took me a while to work out why I couldn't replicate it initially until I found out it was setting MaxConcurrentJobs = 0 which causes the error.

[RQtestapp.zip](https://github.com/tjscience/RoboSharp/files/8109690/RQtestapp.zip)

Call stack also copied below:

>	RoboSharp.dll!RoboSharp.Results.Statistic.AddStatistic(RoboSharp.Interfaces.IStatistic stat) Line 620	C#
 	RoboSharp.dll!RoboSharp.Results.RoboQueueProgressEstimator.Estimator_ValuesUpdated(RoboSharp.Interfaces.IProgressEstimator sender, RoboSharp.EventArgObjects.IProgressEstimatorUpdateEventArgs e) Line 165	C#
 	RoboSharp.dll!RoboSharp.Results.ProgressEstimator.PushUpdate() Line 506	C#
 	RoboSharp.dll!RoboSharp.Results.ProgressEstimator.GetResults() Line 166	C#
 	RoboSharp.dll!RoboSharp.Results.ResultsBuilder.BuildResults(int exitCode, bool IsProgressUpdateEvent) Line 74	C#
 	RoboSharp.dll!RoboSharp.RoboCommand.GetRoboCopyTask.AnonymousMethod__0() Line 530	C#
 	[Resuming Async Method]	
 	[External Code]	
 	RoboSharp.dll!RoboSharp.RoboCommand.GetRoboCopyTask.AnonymousMethod__2(object sender, System.EventArgs args) Line 519	C#
 	[External Code]	
 	[Async Call Stack]	
 	[External Code]	
 	[Async] RoboSharp.dll!RoboSharp.RoboCommand.GetRoboCopyTask.AnonymousMethod__1(System.Threading.Tasks.Task continuation) Line 536	C#
 	[Async] RoboSharp.dll!RoboSharp.RoboCommand.Start_ListOnly(string domain, string username, string password) Line 348	C#
 	[Async] RoboSharp.dll!RoboSharp.RoboQueue.StartJobs.AnonymousMethod__3(System.Threading.Tasks.Task t) Line 627	C#
 	[External Code]	
 	[Async] RoboSharp.dll!RoboSharp.RoboQueue.StartJobs.AnonymousMethod__0() Line 659	C#
 	[External Code]	
 	[Async] RoboSharp.dll!RoboSharp.RoboQueue.StartAll_ListOnly.AnonymousMethod__123_0(System.Threading.Tasks.Task continuation) Line 516	C#
 	[Async] RQtestapp.exe!RQtestapp.Form1.ListOnly() Line 68	Basic

RFBomb:
The error occurred when generating The new event arts object and passing that event or object to the add statistic method.

I thought I had accounted for the possibility of passing Noel statistic objects for the change values into the of an art objects, but maybe I did not. This last commit initializes all of the values to an end value of zero, which maybe will correct the issue. Previously these were never initialized for the static dummies objects

PCAssistSoftware:
> The error occurred when generating The new event arts object and passing that event or object to the add statistic method.
> 
> I thought I had accounted for the possibility of passing Noel statistic objects for the change values into the of an art objects, but maybe I did not. This last commit initializes all of the values to an end value of zero, which maybe will correct the issue. Previously these were never initialized for the static dummies objects

I'm sorry to say but still same error after last commit - and again only if MaxConcurrentJobs is set to 0 - works fine if set to 1

RFBomb:
Try moving`}` on RoboQueue line 162 to above current line 171?

PCAssistSoftware:
> Try moving`}` on RoboQueue line 162 to above current line 171?

Sorry can't see where you mean - no } on line 162 of RoboQueue.cs?

RFBomb:
Sorry, RoboQueueProgressEstimator

PCAssistSoftware:
Moving the } seems to have resolved it.

Out of interest you don't seem to need the previous change i.e.

image

So not sure if that should be kept or not? you will know more than me on that one?

Thank you so much for helping resolve this.

Will continue running more tests throughout the rest of today / tomorrow on this and report back later.

I owe you several beers!!!

RFBomb:
Yea, the change to Statistic initialization was more of a 'throw it at the wall' in hope of fixing the issue than anything. I didn't think it was the issue, but wanted to ensure it. Especially since the actual error was occurring on line 167 or so, but that never showed in the call stack.

PCAssistSoftware:
Unfortunately something still not right here - gone back to just a simple 1 source 1 destination job with /E and not using /MIR

Running **list only** shows correct information in progress estimator for **both** the **job estimate** and **total estimate** - but when running **without list only** it shows correct information for job estimate but wrong information for total estimate

image

RFBomb:
Moving the `}` was simply a test to see if it would resolve that initial issue. 
This last commit cleaned up the locking code to cause less hangs when running multiple jobs.

I also ran a quick test using the source and job options provided in the previous issue you had noted, and both ListOnly and Start() ran without issue in BackupApp. All UI items were identical values at end of the run

PCAssistSoftware:
Just tested with my earlier job information (/E /MIR) and the later one without /MIR and all seems to be running as expected here.

Will continue tidying up all my GUI changes this evening / tomorrow and keep running various tests as I progress, but looks good so far and I much prefer the new `e.ResultsEstimate.ValuesUpdated` event rather than the separate ones we had before, much cleaner to use.

Thanks again for your help

RFBomb:
> I much prefer the new e.ResultsEstimate.ValuesUpdated event rather than the separate ones we had before, much cleaner to use.

Yea, that was what I had originally wanted to do, but with all the other stuff being changed (both here and in my app), I couldn't figure out how to do it cleanly. It took the several iterations and learning how threading interaction works before I finally figured out how to do it. I agree its MUCH cleaner than before.

PCAssistSoftware:
I will continue tidying my code and testing throughtout tonight/tomorrow and if we are both happy tomorrow then can then merge it in

RFBomb:
There is one more tweak I want to try out on this prior to merging it to optimize it, but it will not change the functionality. I simply want to avoid performing the clone if I don't have to, by wrapping it inside of the is non-zero property check of the statistic object

PCAssistSoftware:
Okay, cool - sounds goods

PCAssistSoftware:
Tested, tested and tested again, can't seem to break it this time :) tried with all manner of different job configurations and switches, sources large and small and all progress showing as expected.

RFBomb:
Alright, I have the minor code changes complete on my local repo, but don't have internet currently to upload them. It's a micro-optimization.

 I wrapped the cloning lines into an `if (Statisitic.NonZero)` statement to avoid unnecessary creation of objects whose values are all 0. (The event will use the static dummy objects whose value should never  change off 0 in place of cloning a 0-value object multiple times) 

Once I push this commit tonight, a brief test should show it working fine and it's ready for commit.

PCAssistSoftware:
Okay, I will keep an eye out for the commit and give it a test when it appears and then merge this PR

RFBomb:
Ok, I just uploaded the commits to this PR. I've updated my app to use the ValuesUpdated events and I have to say, its much nicer than updating a progress bar via OnFileProcessed. everything seemed to work properly during my test. It looked smoother and felt less laggy since the updates were periodic (4x per second for roboQueue) instead of instant

PCAssistSoftware:
> Ok, I just uploaded the commits to this PR. I've updated my app to use the ValuesUpdated events and I have to say, its much nicer than updating a progress bar via OnFileProcessed. everything seemed to work properly during my test. It looked smoother and felt less laggy since the updates were periodic (4x per second for roboQueue) instead of instant

Agreed it is much nice using ValuesUpdated to show overall progress information

Just downloaded latest commits and tested but for some reason no longer getting any information for bytes in ValuesUpdated, files and directories working fine? just deleted my source and redownloaded just in case problem my end but same issue - issue occurs in BackupApp and my own code?

PCAssistSoftware:
That's sorted it.

Just ran several more tests and all okay here.

If you are happy then I will merge it.

Yet again thanks for all your great work with this, much appreciated.

RFBomb:
merge away

PCAssistSoftware:
> Ok, I just uploaded the commits to this PR. I've updated my app to use the ValuesUpdated events and I have to say, its much nicer than updating a progress bar via OnFileProcessed. everything seemed to work properly during my test. It looked smoother and felt less laggy since the updates were periodic (4x per second for roboQueue) instead of instant

Hi Robert, 

Can I ask how in your app you handle updating the overall progress using ValuesUpdated events AFTER any errors have occurred e.g. when something like #151 occurs?

I have similar issue with an "ERROR 33 (0x00000021)", "The process cannot access the file because another process has locked a portion of the file." when it comes across in use / locked files, and obviously if this occurs in the middle of a copy job then it completely throws out the results shown in progress estimator and thus those we both then derive from ValuesUpdated to depict overal progress.

In the past (before Progress Estimator and RoboQueue etc) when I was showing calculating/showing overall progress using OnCopyProgressChanged / OnFileProcessed events I used to set a bool to True if OnError occurred, and then in OnFileProcessed I wouldn't calculate that particular files bytes if the bool was True and then set it back to False each time OnCopyProgressChanged event raised - this used to work fine - but obviously things have moved on (in a good way) and this method won't work as events fire differently and at different times.

Just wondered if this was something you had covered in your app or even thought about - don't worry if not - just asking before I devote some time to it this week :)

PCAssistSoftware:
Actually thinking this through further - I think we need to (if at all possible) change how RoboSharp handles errors in relation to progress estimator.

If you run a copy using RoboCopy where the source contains files which are in use then it shows the errors (which in our case would raise the OnError event) it retries as per /R /W switches and then the failed files / dirs are added to FAILED count.  And the same is true with RoboSharp results.

But with Progress Estimator it doesn't show the files as FAILED so throws out all the figures used for overall progress if as above using the ValuesUpdated event to do so.

Thoughts?

RFBomb:
> Can I ask how in your app you handle updating the overall progress using ValuesUpdated events AFTER any errors have occurred e.g. when something like #151 occurs?

So far that's the only time it's erroring. And my app used to use robocopy commands hard codes, typically no issues, so I just wasn't worried about it. 

And that particular error im getting doesn't matter, bcause that's stage 1 of the copy (move files to a backup folder) the. Stage 2 puts mostly identically named files in their place. 

> In the past (before Progress Estimator and RoboQueue etc) when I was showing calculating/showing overall progress using OnCopyProgressChanged / OnFileProcessed events I used to set a bool to True if OnError occurred, and then in OnFileProcessed I wouldn't calculate that particular files bytes if the bool was True and then set it back to False each time OnCopyProgressChanged event raised 

So, I don't understand how it's messing up the calculation, so I don't understand what needs to be updated here. If it errors, it would retry. I believe it shouldn't be counted if it exceeds retry count, and I don't think it does (but can't verify, because I don't know how to cause it to error or retry). 

That said, adding that logic into progress estimator should be doable if you've already figured it out, but I don't know what the actual issue is. 

> this method won't work as events fire differently and at different times.

The ProgressEstimator should be as close as possible to actual, and simply periodically update the UI if desired, instead of overloading it when multiple Statistics objects are all raising their own events. My design intent was for: 
- getting some results if a long job is cancelled
- UI updates in standardized way instead of everyone having to write their own logic to do so (since it's quite a complex logic). 

Again, I'm not sure what the actual issue is, so I don't know how to comment on it, other than share my design goal/intent

RFBomb:
Based on your second comment, it's essentially 'not marking failed files as failed', yes? 

As I noted above, I'm not having files actually fail, so i never tested or evaluated that side of the code. If the failure is explicitly logged (as opposed to just assumed to due multiple fails), then it should be trivial to pass that to the 'AddFile' portion of progress estimator. 

If you want, PE could also be updated to store the system messages and errors. But that was just outside My scope during development.

PCAssistSoftware:
> Based on your second comment, it's essentially 'not marking failed files as failed', yes?

Yes essentially that is the main problem here, and fixing that would solve the first point I was trying to explain I think.

When it comes across files that are in use it reports them in the OnError event, retries X number of times and then moves on to the next file, **BUT** it is **NOT** recording the failed file(s) as FAILED for the progress estimator and it is this which in turn has a knock on effect for my live results I show based on the ValuesUpdated event e.g. where it is counting down files / dirs / bytes based on totals obtained during list only and also updating a progress bar to show overall progress

> As I noted above, I'm not having files actually fail, so i never tested or evaluated that side of the code. If the failure is explicitly logged (as opposed to just assumed to due multiple fails), then it should be trivial to pass that to the 'AddFile' portion of progress estimator.

Example of error showing in OnError:

2022/02/28 00:42:20 ERROR 33 (0x00000021) Copying File C:\Users\darre\AppData\Local\Microsoft\Outlook\filename@domainname .co.uk.ost
The process cannot access the file because another process has locked a portion of the file.

Results correctly show it as two files Failed

image

But progress estimator doesn't

image

> If you want, PE could also be updated to store the system messages and errors. But that was just outside My scope during development.

Not sure if that is needed as just having the name and error code / message in the log / output is probably enough for me / user to know which file(s) failed and why

N.B. Have added a new error code / error description to ApplicationConstants.cs relevant to the error code I was getting

RFBomb:
I'm confused by your pics. 

Is there a total of 4 or 5 files? I assume it's 5 files, and two separate files failed.

Since the 'CurrentFile' is stored via PE, you could do something similar to how 'AddFileCopied' was done for the failed counter. Or set a bool so that the next time it goes to add a file, it registers it as failed (similar to how it currently assumes skipped)

PCAssistSoftware:
its one folder containing 5 files, two of which fail due to being locked - but the fact the confusion about 4 or 5 files proves the point that the locked files breaks the progress estimator!!!

image

RFBomb:
Well, it was an 'estimator'. Without having real time access to RoboCopy's internal counters, every possible scenario would have to be coded in. 

That said, prior to raising OnError, the bool could be set for ProgressEstimator to mark the current file as failed. Try that

PCAssistSoftware:
> Well, it was an 'estimator'. Without having real time access to RoboCopy's internal counters, every possible scenario would have to be coded in.
> 
> That said, prior to raising OnError, the bool could be set for ProgressEstimator to mark the current file as failed. Try that

Oh yes, agreed, I completely appreciate that fact :) and typically an in use file wasn't one I come across during all the testing I had done, until today and after spending all weekend perfecting my app it completely threw me out as I thought I had everything perfected with my overall progress etc until then!

I will do some more testing tomorrow and come up with a test source path to use so that you can see your end as well a failed file and what happens, and then hopefully between us we can come up with a solution as I suspect FAILED files are possibly a common occurance for users of RoboSharp so it would be nice for them to be handled properly if we can with the ProgressEstimator.

Thanks

RFBomb:
What I'm suggesting is that since you already had something figured out previously, based in a bool and OnError, just apply that to PE and RoboCommand so RC sets some bool in PE so PE knows to mark the CurrentFile and failed instead of Skipped (which is default while CopyOperation is true)

PCAssistSoftware:
> What I'm suggesting is that since you already had something figured out previously, based in a bool and OnError, just apply that to PE and RoboCommand so RC sets some bool in PE so PE knows to mark the CurrentFile and failed instead of Skipped (which is default while CopyOperation is true)

That was more based on before we had ProgressEstimator and I had complex methods in OnCopyProgressChanged / OnFileProcessed events to show overall progress but yes I will take a look.

N.B. Ignore confusion over 4 or 5 files in screenshots above - that seems to have been more down to my choice of source to demonstrate locked files (Outlook) than anything else, using another source of a locked file shows that ProgressEstimator wrongly records it as Copied rather than marking it as Failed

RFBomb:
> records it as Copied rather than marking it as Failed

In that case, it's probably occurring when it selects the next file, the previous file does a few checks to see where it should be added (at the top of AddFile and when it ends). So setting a bool as you suggested should work, as it's already what's happening for Copy/Skipped.

PCAssistSoftware:
Okay, I will take a look, but may need some help when you have some time.  My previous implementation using a bool was at GUI level just to work my specific needs, whereas this needs to be implemented within RoboSharp so files / directories raised within OnError are marked as Failed not Copied if possible

PCAssistSoftware:
To simulate an error in RoboSharp where it can't copy a file, shows it in OnError, but marks it as Copied rather than Failed you can use the below PowerShell script.

Just change the path/filename for a text file in line 2, run it and it locks the file until a key is pressed, so you can then run RoboSharp and choose source path containing chosen text file in PowerShell to replicate problem :)

Or if you use Microsoft Outlook then use the folder containing .ost / .pst files as your source whilst Outlook is open.

```powershell
#Specify the file name
$fileName = "C:\Users\darre\Desktop\1a\file.txt"

#Open the file in read only mode, without sharing (I.e., locked as requested)
$file = [System.io.File]::Open($fileName, 'Open', 'Read', 'None')

#Wait in the above (file locked) state until the user presses a key
read-host “Press ENTER to continue...”

#Close the file
$file.Close() 
```

PCAssistSoftware:
Okay, have been looking through the code and think I know what needs to happen but not sure 'how' to implement it, so asking for some help / guidance please.

I am thinking that here we would set a bool such as ErrorOccurred to true

https://github.com/tjscience/RoboSharp/blob/4de9e71e6b3d9bbdb32334a9091027d196674dbf/RoboSharp/RoboCommand.cs#L718-L734

And then here we would check if true if that bool is true and then set WhereToAdd to Failed, and then reset the ErrorOccurred bool back to False?

https://github.com/tjscience/RoboSharp/blob/4de9e71e6b3d9bbdb32334a9091027d196674dbf/RoboSharp/Results/ProgressEstimator.cs#L252

But my knowledge of C# and abilities are not enough to know how to declare a bool which can be used across the project so it can be declared in RoboCommand.cs? and used/seen in ProgressEstimator.cs?  

Or am I completely off the mark here?

RFBomb:
Your thinking is correct. 

Seen line [57](https://github.com/tjscience/RoboSharp/blob/4de9e71e6b3d9bbdb32334a9091027d196674dbf/RoboSharp/Results/ProgressEstimator.cs#L57) for declaring a bool. 

`internal bool FileFailed {get;set;}`

PCAssistSoftware:
> Your thinking is correct.
> 
> Seen line [57](https://github.com/tjscience/RoboSharp/blob/4de9e71e6b3d9bbdb32334a9091027d196674dbf/RoboSharp/Results/ProgressEstimator.cs#L57) for declaring a bool.
> 
> `internal bool FileFailed {get;set;}`

Thanks, at least my logic was good if nothing else.

See PR #155 for what I have done - unfortuantely it doesn't work though as it appears from some quick debugging that AddFile fires **before** OnError occurs so even though I set the bool to true it doesn't matter as AddFile never knows about it?

RFBomb:
AddFile occurs every time RoboCopy switches to a new file in the log. 

The OnError event should raise if that file fails, after which AddFile is called for the next file in line. 

Your edit in PE is looking at the newly submitted file instead of the previous one. (See the If skipped statement - capitalization is important.) 

Also, a file cannot be skipped and failed, so it needs to go into an ElseIf chain - if failed else if skipped.

Finally, there is another location where 'if skipping' is evaluated at the end of the run. This will need to be checked there as well

PCAssistSoftware:
I have two files in a folder, one which is locked and one which can be copied fine.  

I put breakpoints on the following two lines:

A) Line 255 of ProgressEstimator.cs            AddFile

image

B) Line 719 of RoboCommand.cs                OnError

image

The breakpoints are hit in this order:

    A (for the first file)   
    A (for the second file)   
    B (to raise the error for the file it can't copy)

It DOESN'T raise the AddFile event again after raising the error event? 

> Your edit in PE is looking at the newly submitted file instead of the previous one. (See the If skipped statement - capitalization is important.)

Yes I noticed the captilization and tried that too - but as above after hitting second breakpoint e.g. when raising error it doesn't then raise AddFile again, so it doesn't see I have set bool to True etc.

RFBomb:
https://github.com/tjscience/RoboSharp/blob/4de9e71e6b3d9bbdb32334a9091027d196674dbf/RoboSharp/Results/ProgressEstimator.cs#L171

PCAssistSoftware:
> https://github.com/tjscience/RoboSharp/blob/4de9e71e6b3d9bbdb32334a9091027d196674dbf/RoboSharp/Results/ProgressEstimator.cs#L171

Just placed a breakpoint here and yes it gets to that point but then jumps straight to the else if because CopyOpStarted = False, so adds it to skipped rather than failed

RFBomb:
Right, because your not checking for the failed bool

PCAssistSoftware:
Forget my changes for a minute, it is getting to that line you mentioned anyway for the file which is generating the error so isn't that line supposed to be dealing with it anyway?  What is that section for?  It looks like it is creating the RoboCopyResults to me? rather than the estimator values?

Even more confused now.   Will look again later with a clear head

RFBomb:
I wrote remarks at the top of the method describing its usage.

PCAssistSoftware:
> I wrote remarks at the top of the method describing its usage.

Yes I appreciate that and read them but still can't get it working, tried adding in various different bits of code to that section to check my bool and then process the file as failed but still doesn't show as failed in progress estimator

RFBomb:
I haven't tested anything, but I suspect something like this should work:

if ((FileFailed | CopyOpStarted) && CurrentFile != null)

PCAssistSoftware:
Doesn't seem too unfortunately :(

IGNORE THIS - SEE BELOW

PCAssistSoftware:
Update: actually putting in a breakpoint and checking the variables I can see the values **are** now updating with the correct values :)  but they are just not being shown in the GUI after the operation is complete

image

PCAssistSoftware:
Update2: still issues as tried with more files and sometimes my breakpoint at the changed 171 is not even hit despite error being raised and file failing to copy, so confused now and need to look again with clear head tomorrow - have pushed all the last tested changes to the PR, as last update it worked fine with just a couple of files (just wasn't updating GUI, but figures showing okay in breakpoint), but with more files something not working - more tomorrow

PCAssistSoftware:
Okay to explain above further before I retire for the evening

First tests were just with 2 files - one locked - one which copies okay - breakpoint showed it working okay, only issue was just not updating GUI to show the failed results.

Second test was with 3 files - one locked - two which copy okay - it seems with more files the locked file is now being reported by progress estimator as skipped ?! (can see by the size - 6 bytes) - and the breakpoint at line 171 is not now hit as when it reaches it CurrentFile is null

image

image

RFBomb:
Revert https://github.com/tjscience/RoboSharp/commit/317fe641bc1eb74c438d0c3e43579c1c0f304917

Not sure why you removed the changes to AddFile that evaluate the bool

PCAssistSoftware:
> Revert [317fe64](https://github.com/tjscience/RoboSharp/commit/317fe641bc1eb74c438d0c3e43579c1c0f304917)
> 
> Not sure why you removed the changes to AddFile that evaluate the bool

Whoops, don't know what happened there, didn't meant to remove it!

Anyway with the three file test it now works fine and it seems to update the GUI okay

BUT if I go back to my original two files and the last file happens to be the one which fails then the progress estimator doesn't then refresh to show the final results - so any way we can call one last refresh so it shows the correct information?

image

if I rename the first file so alphabetically it appears **after** the locked file then GUI updates successfully, so issue only occurs if last file processed just happens to be a FAILED file 

image

RFBomb:
Good catch on it not doing a last update. 

Move line 168 `PushUpdate(); // Perform Final calculation before generated Results Object
`
To below the if/else lines for skipped/failed. 

Convert the lines inside of Skipped/Failed to their PerformByteCalc() equivalents. 

This should cause it to perform that last update.

PCAssistSoftware:
That seems to work, thank you for helping me sort it.

Does it all look okay to you?

Will test it tomorrow with my GUI as so far just been testing with BackupApp

Does FileFailed = false; need to be in both places?

RFBomb:
I'll take a look this week when I get a chance

Issue #153, opened 2022-02-22
Method to save multiple jobs (e.g. content of RoboQueue)


PCAssistSoftware:
We currently have a great way to save single jobs (.RCJ files) from RoboSharp which works really well (thanks @RFBomb)

But it would be nice to also have a way to be able to save the contents of a "RoboQueue" e.g. all the jobs you currently have in your queue and their related settings.

**My Initial Thoughts**

- Preferably to a single file if possible, rather than lots of separate ".RCJ" job files, as it could get quite messy if you have a lot of jobs in your queue.

- Unlike single jobs these don't necessarily need to be backward compatible with RoboCopy so we can use whatever format we like? perhaps XML would be a good fit? or JSON?

Just opening it up here for discussion, with the hope that someone may be able to help implement something in the future :)

Always here for testing and inspiration....... and whatever other help I can provide......

RFBomb:
As far as readability goes, my personal take is that XML is the way to go, but that's what I'm used to. 

I would suggest trying to simply use the C# xml serializer to see what it spits out into a file when run against the RoboCommand/RoboQueue objects. 

You can tag properties with [XMLIgnore] to avoid the serialize caring about them. Also note only public properties with Get and Set will be serialized.

PCAssistSoftware:
Been a while since I last looked at this, but now have a need for it again, so finally got round to looking at it again.

In hindsight actually not sure if I really need to save the contents of a "RoboQueue" as first thought, but I do need a way to save the individual RoboCommand's from a RoboQueue so to speak.   So for example if I have added several jobs to a queue, to either be able to loop through the queue and save each one to a .RCJ file, or to select a specific one (by name) and save that.

BEFORE adding the RoboCommand's to the queue I could easily use the SaveasJobFile Function.  But AFTER they are in a RoboQueue there is no easy way that I can see that I can then do this.  It would be handy if you could loop through all jobs in a queue and save them or just select one by name and save it.  Would this be a plausible idea?     Hope this make sense, as typing it late at night as I think it through.

PCAssistSoftware:
Once you have used "SaveAsJobFile" how do you then get it back in again - as can't find a corresponding "LoadFromJobFile"??

RFBomb:
That's a good idea for a static RoboCommand method, that returns a RoboCommand. 

I just took a quick peek to refresh myself though. Since the JobFile has to be parsed, and isn't guaranteed to contain all info related to running a job. 

[...]

PR #154, opened 2022-02-23
XML Writer for RoboCommand and RoboQueue

RFBomb:
@PCAssistSoftware - Regarding #153 - Since I've worked with the xml serializer in the past, I wrote up this very basic class to serialize RoboCommand and RoboQueue.

Theres a lot more work that needs to be done before this would get released, but I wanted to provide you a starting point to work on it. ( I can review it and chime in, but would rather not dedicate myself to this, as I have other pressing projects, and you are the one primarily requesting it.)

Currently, I had to tag the `CopyOptions.FileFilters` with `[XMLIgnore()]` to get it to serialize robocomand. I have not written anything for loading it yet. I also simply mocked in the xml writer call into the current 'Save Job' method within robocommand, which is then wrapped in `#IF DEBUG` to avoid accidentally releasing it. 

In Visual studio the 'Task List' can be seen on things to do (Duplicates show up due to multiple framework targets, so I just filtered down to Net45)

image

Here is the file that RoboCommand serializer spit out. It should deserialize fine, but I haven't tested it. 
[Test.zip](https://github.com/tjscience/RoboSharp/files/8124384/Test.zip)

Also, methods for RoboQueue will need to be written and tested.

### DIFF
--- a/RoboSharp/CopyOptions.cs
+++ b/RoboSharp/CopyOptions.cs
@@ -139,8 +139,10 @@ namespace RoboSharp
         /// Allows you to supply a set of files to copy or use wildcard characters (* or ?). <br/>
         /// JobOptions file saves these into the /IF (Include Files) section
         /// </summary>
+        [System.Xml.Serialization.XmlIgnore()]
         public IEnumerable<string> FileFilter
         {
+            //TODO: Modify to allow serialization
             get
             {
                 return fileFilter;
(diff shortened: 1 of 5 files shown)


PCAssistSoftware:
Excellent - thank you for this, when I have a bit more time myself I will spend a while getting my head around it and see what I can do :) - much appreciated

RFBomb:
Superceded by #183

PR #155, opened 2022-03-01
attempts to fix errors not showing as failed in progress estimator

PCAssistSoftware:
attempts to fix errors not showing as failed in progress estimator

### DIFF
--- a/RoboSharp.BackupApp/MainWindow.xaml
+++ b/RoboSharp.BackupApp/MainWindow.xaml
@@ -4,7 +4,7 @@
         xmlns:av="http://schemas.microsoft.com/expression/blend/2008" xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" xmlns:local="clr-namespace:RoboSharp.BackupApp" mc:Ignorable="av" x:Class="RoboSharp.BackupApp.MainWindow"
         Title="Backup App" Height="700" Width="750" ResizeMode="CanResizeWithGrip" MinWidth="750" MaxWidth="750" MinHeight="650" >
     <Grid IsSharedSizeScope="True">
-        <TabControl SelectionChanged="TabControl_SelectionChanged">
+        <TabControl>
             <TabItem Header="Options" x:Name="OptionsTab">
                 <Grid>
                     <ScrollViewer Margin="0,0,0,0" VerticalScrollBarVisibility="Auto" Width="Auto" Height="Auto">
@@ -114,7 +114,7 @@
                                         <CheckBox Grid.Row="6" Grid.Column="2" x:Name="EnableRestartModeWithBackupFallback" Margin="10,0,0,0" />
                                         <CheckBox Grid.Row="7" Grid.Column="2" x:Name="UseUnbufferedIo" Margin="10,0,0,0" />
                                         <CheckBox Grid.Row="8" Grid.Column="2" x:Name="EnableEfsRawMode" Margin="10,0,0,0" />
-                                        <TextBox Grid.Row="9" Grid.Column="2" x:Name="CopyFlags" Margin="10,0,0,0" PreviewTextInput="IsAttribute_PreviewTextInput" Width="100" HorizontalAlignment="Left" />
+                                        <TextBox Grid.Row="9" Grid.Column="2" x:Name="CopyFlags" Margin="10,0,0,0" PreviewTextInput="IsCopyFlag_PreviewTextInput" Width="100" HorizontalAlignment="Left" />
                                         <CheckBox Grid.Row="10" Grid.Column="2" x:Name="CopyFilesWithSecurity" Margin="10,0,0,0" />
                                         <CheckBox Grid.Row="11" Grid.Column="2" x:Name="CopyAll" Margin="10,0,0,0" />
                                         <CheckBox Grid.Row="12" Grid.Column="2" x:Name="RemoveFileInformation" Margin="10,0,0,0" />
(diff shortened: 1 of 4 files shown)


RFBomb:
This looks good to me. 
To answer your other question about if thats where the 'FileFailed = false' lines should be, those locations were fine, but to keep in practice with where I had placed the other flag resets, I moved them a bit.

PCAssistSoftware:
Thanks Robert, really appreciate you looking it over for me.

Before reading the next sentence please pour yourself a drink......... ;)

Just one issue I have found and yes it has to do with my really annoying 0kb files again (sorry I couldn't resist trying them!!) - for some reason if the file that is raising OnError happens to be a 0kb file then this fix doesn't work and instead of being shown as FAILED it is shown as COPIED

Any thoughts?  Not a huge worry as the likelihood of it happening I suppose is quite slim but as I had tested everything else with same set of source files I had to just try it.

PCAssistSoftware:
> Before reading the next sentence please pour yourself a drink......... ;)

P.S. And that drinks on me :)

RFBomb:
0kb files are added immediately during AddFile, prior to error being caught. 

Only way around this would be deferring their addition until the next file is added, but that's gets confusing due to it being a special case. 

'CopyOpStarted' could be set true in this case, and another if statement that checks that AND that the CurrentFile.Size == 0 (to avoid double counting) to perform the calc. 

I can put this in in the morning, it will be quick

PCAssistSoftware:
Thanks Robert (sorry for delay in replying)

RFBomb:
Those stupid 0kb files are a nuisance. 
This should resolve it, but I'm done working around them

RFBomb:
make sure to retest everything regarding the output from ProgressEstimator as the AddFile has been completed restructured

PCAssistSoftware:
> Those stupid 0kb files are a nuisance. This should resolve it, but I'm done working around them

Thank you very much, and I completely understand what you mean!! 

Let me know what you drink and where you live (my contact details are in my profile) and I will come up with some form of suitable thank you :) as I do really appreciate everything you have done to help me

PCAssistSoftware:
> make sure to retest everything regarding the output from ProgressEstimator as the AddFile has been completed restructured

Will run through all my original tests again over the weekend to fully ensure that everything is as it shoud be with the Progress Estimator and will let you know

PCAssistSoftware:
Have spent last hour going through dozens of different combinations of sources and switches and all seems to be working perfectly fine.

Will do some more testing tomorrow with some larger sources just to mak sure.

Thanks again.

RFBomb:
What would be really smart to do for future testing here would be build a UnitTesting project included here. 

In the unit test solution have a folder tree with a bunch of small files (0-2kb). 

Then have the unit test generate and run commands, then compare the results of ProgressEstimator to actual results. 

Doing so would be fairly complex, but would prevent us running tests manually over and over. 

I've never built a unit testing service like this, but essentially it would probably be relatively simple to do.

Static methods for:
- Generating the base command with source/destination
- method to run the test in a normalized way and compare the results, handing them back to the unit test being run
- Method to purge the test destination 


Then each UNIT TEST will simple call those static methods, turning options on/off as needed for the test. 

(Since it will be housed in its own project, it could be any language)

PCAssistSoftware:
Yes that would make sense - unit testing is not something I have done either, but will add it to my "TO DO" list as would certainly be worth it, and an area I would like to learn, and would certainly be less time consuming than manually running lots of individual tests each time.

PCAssistSoftware:
Have created an issue (labelled as discussion / enchancement) #156 covering your great idea of adding an unit testing project in the future - quite like idea of looking into this sometime in future when time allows as would like to learn more about doing unit testing

PCAssistSoftware:
@RFBomb 

I have done as much testing as I can with these changes and as far as I can see all results for ProgressEstimator are appearing correctly for every single combination of source and job option I can think of.  Thanks yet again for sorting this.

I also noticed during testing that the check in the GUI for flags you could enter for attribute exclusion / inclusion was missing some such as R for read-only so I have fixed that (IsAttribute_PreviewTextInput), I also noticed same check was being used for CopyFlag even though flag values are completely different so to save confusion and wrong flags being used / entered I have added new check for IsCopyFlag_PreviewTextInput

Unless you have anything to add / change, and if you agree then I will merge the changes soon.

PCAssistSoftware:
Separate to this PR but I wonder whether at some point it would be possible to include the error information in the results object - as would be much easier to work with here.

For example at the moment when a file error occurs we just get / store a string of information containing the date/time, file name, error code and error message e.g. "2022/03/05 16:22:54 ERROR 2 (0x00000002) Changing File Attributes C:\Users\darre\Desktop\1c\zzz.txt" and it is hard to do much with this other than just showing it to the user as it appears.

Would be nice to retrieve each element separately for each error which occurs e.g. error code, error message, file name, file size etc, so I can then easily show a message or bind the information to a grid etc, or to show detailed error information in reports / logs.

Just a thought

RFBomb:
You should be able to accomplish this relatively easily.
Add a list or error objects to the ResultsBuilder object, then when  'FileFailed = true' occurs, add it to the list before raising the event.
THen just send that list to the results object once its complete. (List should be returned as an array to help avoid further modification)

PCAssistSoftware:
Okay thanks, I will have a go - I will do it under a separate PR though to keep things clean - are you happy that this PR can be merged?

RFBomb:
If all was working, I'm ok with it

PR #157, opened 2022-03-05
Add error information to RoboCopyResults / RoboQueueResults

PCAssistSoftware:
Add information from OnError event to RoboCopyResults / RoboQueueResults so that we can easily access details such as file name / path, file size(?), error code and error message for each file that failed to copy.


PCAssistSoftware:
@RFBomb 

> You should be able to accomplish this relatively easily.
> Add a list of error objects to the ResultsBuilder object, then when 'FileFailed = true' occurs, add it to the list before raising the event.
> Then just send that list to the results object once its complete. (List should be returned as an array to help avoid further modification)

Draft PR created - when you have some time (no hurry whatsoever) I would really appreciate some pointers as to what needs adding and where to achieve this if possible.  Really not sure where to even start and don't want to break anything or over complicate it.

PCAssistSoftware:
N.B. I had initially looked at just handling it myself in my own GUI but unfortuantely very limited information available in the OnError event as all you get is a preformatted string containing date/time (in US format), file name, and error code/description all together, so to get useful information I would have to parse it to split it into date/time (and covert it to UK), file name, error code and error description - so thought perhaps better just to have information separated and available in the results in first place in RoboSharp rather than the other way round.

RFBomb:
Sounds like you should be focused on modifying the Error event object. 

RoboSharp/EventArgObjects/ErrorEventArgs.cs

Modify that file so that when it is created the constructors parses the error string and builds the properties you want to add. 

For example, it can parse the date from the string and store it as a DateTime property. 

As for having this in the results object:

Add List<ErrorEventArgs> property to the results builder and prior to raising the event create the object and add it to the list. 

The robocopyresults object (and its interface) should have an ErrorEventArgs[]` property.

PCAssistSoftware:
Yes I have been looking at that today.

Not really interested in date - as not really relevant - only mentioned it as the one currently stored is in US format only which is a pain, but as I say not really relevant.

I just want the file name / path, file size and error code / description so I can access each element separately for my logging / reporting.

Not sure how to do what you say for the results builder but will continue playing, as I have said before I program in VB, so working in C# makes everything so much more confusing to work with especially not knowing the code as well as you clearly do :)

Will continue playing though and see what I can come up with

PCAssistSoftware:
I can see how to modify ErrorEventArgs.cs to add the properties I want to add, that part I am fine with.

But I can't see how in I can access the information I need https://github.com/tjscience/RoboSharp/blob/a87a13645cfe60e193a08d66b4fd437b34cd9311/RoboSharp/RoboCommand.cs#L718-L740 

I assume I have to parse the content of the `data` string somehow (RegEx?) to pull out the file name\path? but not sure how I would then get size of the file (although actually thinking it through further perhaps size isn't that important and just have file name/path would be enough)

RFBomb:
So a new regex can be used to break the error code into its various segments, or string.Split() can be used. 

That would be my approach, since it allows more test cases than ( testing if a filename is included in the string for example) and using named groups. I can whip a quick one up if you provide me a few errror codes to base it off f of

RFBomb:
My other question is: 

- can this event be raised without a file being in referenced? 
- can it be raised for a directory?
- can it be raised with no reference to a directory or file?

PCAssistSoftware:
> So a new regex can be used to break the error code into its various segments, or string.Split() can be used.
> 
> That would be my approach, since it allows more test cases than ( testing if a filename is included in the string for example) and using named groups. I can whip a quick one up if you provide me a few errror codes to base it off f of

Here a couple of examples of what the `data` string contains at that point

"2022/03/06 18:07:17 ERROR 2 (0x00000002) Changing File Attributes C:\\Users\\darre\\Desktop\\1c\\file.txt"

"2022/03/06 18:08:28 ERROR 33 (0x00000021) Copying File C:\\Users\\darre\\AppData\\Local\\Microsoft\\Outlook\\darren.rose@myemaildomain.co.uk.nst"

PCAssistSoftware:
> My other question is:
> 
> * can this event be raised without a file being in referenced?
> * can it be raised for a directory?
> * can it be raised with no reference to a directory or file?

never seen it raised myself without a file being referenced - but would be interesting to see what is in the `data` string at line 718 for you when you get error as per your issue #151 as that may answer these questions

Mgamerz:
Does this regex work in other OS languages?

RFBomb:
@PCAssistSoftware, please try this out. See the 'rework' notes I put into that commit's description. Hopefully you can understand what my thought process was here when I recomended using regex to split it into the properties. 

A significant change I also did was delay raising the event until the following command is entered into the log. This is because from everything I've seen, RoboCopy reports errors onto 2 seperate lines, the error, then the description. So the error is processed to mark the file as failed, but the even is raised on the next line that gets processed, so it can report the description by as reported by robocopy (instead of hard coding them in English into ApplicationConstants dictionary)

RFBomb:
@Mgamerz - I only work in English, so I don't know. But in theory it should work fine. If you have another language you want supported, I would highly suggest opening up a Pull Request for it. The framework is already in place for most all the reactions to robocopy (I think). You would require settings up a new DefaultConfiguration class, seen here:

https://github.com/tjscience/RoboSharp/blob/a87a13645cfe60e193a08d66b4fd437b34cd9311/RoboSharp/DefaultConfigurations/RoboSharpConfig_EN.cs

You would modify the values of all the tokens, and if needed modify the regex (or use the regex generator method added in my last commit) to be able to pull out the values from the lines reported by robocopy.

PCAssistSoftware:
> @PCAssistSoftware, please try this out. See the 'rework' notes I put into that commit's description. Hopefully you can understand what my thought process was here when I recomended using regex to split it into the properties.

Currently getting a System.NullReferenceException: 'Object reference not set to an instance of an object.' on line 115 of RoboSharpConfiguration.cs?

image

Yes I fully get your thought process re using the regex to split the properties

> A significant change I also did was delay raising the event until the following command is entered into the log. This is because from everything I've seen, RoboCopy reports errors onto 2 seperate lines, the error, then the description. So the error is processed to mark the file as failed, but the even is raised on the next line that gets processed, so it can report the description by as reported by robocopy (instead of hard coding them in English into ApplicationConstants dictionary)

That also makes sense and is definitely an improvement

PCAssistSoftware:
It now runs without previous error I mentioned above (System.NullReferenceException), but the OnError event isn't being raised at all now, so error not showing in GUI or being caught as FAILED by Progress Estimator?

> A significant change I also did was delay raising the event until the following command is entered into the log. This is because from everything I've seen, RoboCopy reports errors onto 2 seperate lines, the error, then the description. So the error is processed to mark the file as failed, but the event is raised on the next line that gets processed, so it can report the description by as reported by robocopy (instead of hard coding them in English into ApplicationConstants dictionary)

Can't comment for how RoboCopy creates it output, but for RoboSharp the error being spread over two lines I assumed was just because of the Environment.NewLine being included in https://github.com/tjscience/RoboSharp/blob/a87a13645cfe60e193a08d66b4fd437b34cd9311/RoboSharp/RoboCommand.cs#L734

RFBomb:
Robosharp was doing it by performing a lookup of known error descriptions, then applying the appropriate one if found with a new line insert eyed between the two. Then the next Line processed by the log would be the description, fail all the if statements do nothing but adding to log would occur. 

I hoped this change would avoid us having to create that lookup dictionary. I'll test it tomorrow

RFBomb:
Correct me if I'm wrong, it's currently not raising the OnError, correct?

PCAssistSoftware:
> Correct me if I'm wrong, it's currently not raising the OnError, correct?

Yes that's correct - it is NOT raising the OnError so errors not showing in GUI and also errors not being caught as FAILED by Progress Estimator

RFBomb:
I'll take a look, I'm getting on a flight now, so I'll be radio silent for a bit. But if If i figure it out before I land I'll have a few commits for both PRs. 

I added a `public virtual bool ListOnly => false` to the top of the class for ProgressEstimatorTests, and all the GenerateCommand() calls now end with , `ListOnly);` to utilize the property.

Finally, the generate command itself sets list only mode based on that bool that is passed in. 

It might be easier to hold off on working on it till o can push the commit, but if you wanted to get a start on it those are the changes that will eventually need to be done to the tests

PCAssistSoftware:
That's fine, I will have another commit later to add FileInclusion.  Need to think about tests for MaxAge / MinAge as to whether even worth adding (I never use them day to day, only used them when testing switches), but had to use a tool (https://www.nirsoft.net/utils/bulk_file_changer.html) to change dates of files as couldn't easily do it from within Windows or with VS, and I think even if we change the date / time on a file in the TestFiles folder it would potentially change it when it downloads the files on the users computers or when it copies them to bin folder during build etc, so may just make whole test confusing

RFBomb:
The nice thing here is that we simply just have to over the next few days re-save the files in the test source to update their dates. 

We may be able to programmatically change them though to force the test.

PCAssistSoftware:
> The nice thing here is that we simply just have to over the next few days re-save the files in the test source to update their dates.
> 
> We may be able to programmatically change them though to force the test.

Yes, just found these which should work perfectly

https://docs.microsoft.com/en-us/dotnet/api/system.io.file.setcreationtime?view=net-6.0
https://docs.microsoft.com/en-us/dotnet/api/system.io.file.setlastaccesstime?view=net-6.0
https://docs.microsoft.com/en-us/dotnet/api/system.io.file.setlastwritetime?view=net-6.0

RFBomb:
Alright, just landed. I didn't get a chance to look at the OnError event, but I merged your unit test commit and mine, and updated it to iterate through all the RASHCNETO tests for include and exclude. 3 are failing, I don't think we can be force the 'compressed', 'encrypted' attributes, but before I push it I think I have a fix for the 'normal' one.

PCAssistSoftware:
> Alright, just landed. I didn't get a chance to look at the OnError event, but I merged your unit test commit and mine, and updated it to iterate through all the RASHCNETO tests for include and exclude. 3 are failing, I don't think we can be force the 'compressed', 'encrypted' attributes, but before I push it I think I have a fix for the 'normal' one.

I have done Include Attribute and Max Last Access Date tests - is it okay for me to push my changes now?

PCAssistSoftware:
> Alright, just landed. I didn't get a chance to look at the OnError event, but I merged your unit test commit and mine, and updated it to iterate through all the RASHCNETO tests for include and exclude. 3 are failing, I don't think we can be force the 'compressed', 'encrypted' attributes, but before I push it I think I have a fix for the 'normal' one.

I did wonder about the 'additional' attributes

RFBomb:
The Regex used for the error token didn't have the "IDENTIFIER" string replaced with the token. 
Fixed and passing all unit tests (including the ones I just added that verify all events are firing for RoboCommand and RoboQueue)

Can you take a look at the ErrorEventArgs property and see if its up to spec? All parts of the error have been parsed out using the regex, and applied as properties of the object. Still on your to do would be the applying it to the results in a `ErrorEventArgs[] ` object

PCAssistSoftware:
> The Regex used for the error token didn't have the "IDENTIFIER" string replaced with the token. Fixed and passing all unit tests (including the ones I just added that verify all events are firing for RoboCommand and RoboQueue)

Great work on all the new unit tests, this must cover just about every possibilty now

> Can you take a look at the ErrorEventArgs property and see if its up to spec? All parts of the error have been parsed out using the regex, and applied as properties of the object. 

All looking really good to me, and exactly what I was hoping for :)

>Still on your to do would be the applying it to the results in a `ErrorEventArgs[] ` object

I would (please) need some input / pointers from you as to exactly how I go about doing that though?

RFBomb:
It was easier to show than explain. Review all the code changes in the last commit to see what was required. 
While I was at it, I put it in the various interfaces

RFBomb:
Ok, I found why the 'Appveyor' keeps failing. I have no idea WHY though. 
Its failing on the unit tests that performs the FileLocked operations. The ones that generate the errors. 
When you look at the log lines and compare them to tests run from Visual Studio, the logs on Github have no error at all in them!
Its almost like the file is not being held in a locked state. I've tried multiple types of file locks File.Read(), File.Write(), File.Open() and none pass the test on github, but all pass the test locally. 

Also noteworthy, the flags for the DCOPY option differ between local and github, but I don't think that should matter for the file.

PCAssistSoftware:
Sorry got caught up this evening and it is now late here in UK - I can see you have been busy (and thank you very much indeed for adding the errors to results), I have a clear day tomorrow so will catch up on all the commits and reply in more detail then.

PCAssistSoftware:
> It was easier to show than explain. Review all the code changes in the last commit to see what was required. While I was at it, I put it in the various interfaces

Thank you so much for doing this, I have spent some time looking through the code changes and thought I can understand what you did I certainly don't think I would have been able to do it myself, so much appreciated.

I am now able to do exactly what I wanted to do in my logs / report / GUI.

IGNORE BELOW - CAUSED MY OWN CONFUSION AS PER LATER RESPONSE.....

One related thing I have now noticed - when running a job which completes with errors the status at the end has **Successful** set as **True** and **HasWarnings** set as **True**, but **HasErrors** set as **False**? surely should be **HasErrors** as **True** and the others two as False?

image

PCAssistSoftware:
> Ok, I found why the 'Appveyor' keeps failing. I have no idea WHY though. Its failing on the unit tests that performs the FileLocked operations. The ones that generate the errors. When you look at the log lines and compare them to tests run from Visual Studio, the logs on Github have no error at all in them! Its almost like the file is not being held in a locked state. I've tried multiple types of file locks File.Read(), File.Write(), File.Open() and none pass the test on github, but all pass the test locally.
> 
> Also noteworthy, the flags for the DCOPY option differ between local and github, but I don't think that should matter for the file.

This is all very new to me, I never actually realized that once you added unit tests that GitHub / Appveyor actually ran the tests as part of the build checks - very clever, but possibly quite annoying when we know the tests work perfectly well locally, and I suspect in a lot of projects not all unit tests (especially for more complex projects) would work within Appveyor even if working locally, especially if they have external requisites.  

Is there anyway to suppress warnings for certain parts of the code for appveyor in the same way you can do for warnings in code for Visual Studio?

PCAssistSoftware:
IGNORE BELOW - CAUSED MY OWN CONFUSION AS PER LATER RESPONSE.....

@RFBomb 

Further to my last response

> One related thing I have now noticed - when running a job which completes with errors the status at the end has Successful set as True and HasWarnings set as True, but HasErrors set as False? surely should be HasErrors as True and the others two as False?

For my job with errors above the exitCodeValue I receive is 10

In RoboCopyResultsStatus.cs the exitCodeValue is being treated as an integer on lines 21 and 23 but then on lines 33, 36, 39 and 42 the bool check seems to be using hexadecimal / base16 e.g. 0x10 to check for 10 which obviously fails

Therefore the check below when passed an ExitCodeValue (integer) of 10 is True which is wrong

`public bool Successful => !WasCancelled && ExitCodeValue < 0x10;`

As it should surely be

`public bool Successful => !WasCancelled && ExitCodeValue < 10;`

Which returns the correct bool of False

Same for the other checks, I have amended the code for testing and I then get the desired results e.g. HasErrors is now True and Successful is False for my above test - does the below change look correct to you? any comments / thoughts?

 ```
 public RoboCopyExitStatus(int exitCodeValue)
 {
     ExitCodeValue = exitCodeValue;
 }

 /// <summary>ExitCode as reported by RoboCopy</summary>
 public int ExitCodeValue { get; protected set; }

 /// <summary>ExitCode reported by RoboCopy converted into the Enum</summary>
 public RoboCopyExitCodes ExitCode => (RoboCopyExitCodes)ExitCodeValue;

 /// <inheritdoc cref="RoboCopyExitCodes.FilesCopiedSuccessful"/>
 public bool Successful => !WasCancelled && ExitCodeValue < 10;

 /// <inheritdoc cref="RoboCopyExitCodes.MismatchedDirectoriesDetected"/>
 public bool HasWarnings => ExitCodeValue >= 4 && ExitCodeValue < 10;

 /// <inheritdoc cref="RoboCopyExitCodes.SeriousErrorOccurred"/>
 public bool HasErrors => ExitCodeValue >= 10;

 /// <inheritdoc cref="RoboCopyExitCodes.Cancelled"/>
 public virtual bool WasCancelled => ExitCodeValue < 0;
```

Also surely you wouldn't have both HasWarnings and HasErrors both being True? it would be one or the other depending on the seriousness? hence I have amended check for HasWarnings to be more than or equal to 4 or less than 10, as HasErrors is more than 10.  Again thoughts please?

RFBomb:
Idk man, I'm not using these in my project currently. Best I can say is they should probably all be evaluating the exit code, and should be verified to match the robocopy exit code documentation. 

It's probably best to just convert it to the exit code Enum the use `enum.HasFlag()` method to set these bools

RFBomb:
If you feel like it, these could be additional unit tests. I'd put them in their own file mostly for organization purposes if you wrote them. 

As for the AppVeyor failing. I don't think it's something we can suppress. Their version of robocopy is simply not reporting errors into the log, which is why they fail here and not locally. I've opened up 2 support threads on it (GitHub community and on help.AppVeyor )

- Only thing I can think of is they have a bad or old version of robocopy. Maybe we can package a known working robocopy executable  into the unit test project, unzip it and use that as the executable for the tests.  It it's a bit of work. Pus the way the tests are now allows easier migration into other languages because it could be used to force an error, thus allowing us to take that log and generate THe tokens for that language. I'll try it out anyway though.

PCAssistSoftware:
> As for the AppVeyor failing. I don't think it's something we can suppress. Their version of robocopy is simply not reporting errors into the log, which is why they fail here and not locally. I've opened up 2 support threads on it (GitHub community and on help.AppVeyor )
> 
> * Only thing I can think of is they have a bad or old version of robocopy. Maybe we can package a known working robocopy executable  into the unit test project, unzip it and use that as the executable for the tests.  It it's a bit of work. Pus the way the tests are now allows easier migration into other languages because it could be used to force an error, thus allowing us to take that log and generate THe tokens for that language. I'll try it out anyway though.

As you say might be worth a try to see if the results are different, but I suspect as I said in previous reply this is a common problem with different results between what results you get running unit tests locally and what you get with AppVeyor as so much can depend on versions of executables, what file system it is being run on, OS versions etc

PCAssistSoftware:
> Idk man, I'm not using these in my project currently. Best I can say is they should probably all be evaluating the exit code, and should be verified to match the robocopy exit code documentation.
> 
> It's probably best to just convert it to the exit code Enum the use `enum.HasFlag()` method to set these bools

REPLY DELETED AS I CAUSED MY OWN CONFUSION.......

PCAssistSoftware:
Update: I think actually part of this may just be poor wording choice for the HasErrors boolean / combined with my misunderstanding - as I think originally it was just designed to be used only if in the case of Serious Errors:-

> 0×10  16       Serious error. Robocopy did not copy any files.
>                    Either a usage error or an error due to insufficient access privileges
>                    on the source or destination directories.

Rather than just "Errors" e.g. files which didn't copy because of errors - these are just classed as "HasWarnings"

So it is probably behaving correctly and it is just me confusing matters, again....

Ignore me......

Anyway back to original point of this PR, which thanks to you is resolved and works really well, I now have the error information I wanted in the results, so thank you very much indeed.

Do you want to continue trying bits in this PR relating to appveyor before I merge it?   quite happy to do so if you wish - no hurry on my part

RFBomb:
Regarding the Enum, it's pretty easy to do, since it's already tagged with the 'flags' attribute. 

As long as`ExitCode` is the proper value on line 30 here: RoboCopyResultsStatus.cs

Then each of the bool properties can use

`=> ExitCode.HasFlag(RoboCopyExitCodes.FlagOfChoice);`

But you are right that it's currently passing everything as an int32, whereas all those codes refer to hex. Which may be a problem, but I don't believe it is.

RFBomb:
> Do you want to continue trying bits in this PR relating to appveyor before I merge it? quite happy to do so if you wish - no hurry on my part

We know the tests pass. I can play with my repo to see if it's worth doing a pr on the AppVeyor stuff. 

If you are satisfied here, merge away.

PCAssistSoftware:
> Regarding the Enum, it's pretty easy to do, since it's already tagged with the 'flags' attribute.
> 
> As long as`ExitCode` is the proper value on line 30 here: RoboCopyResultsStatus.cs
> 
> Then each of the bool properties can use
> 
> `=> ExitCode.HasFlag(RoboCopyExitCodes.FlagOfChoice);`
> 
> But you are right that it's currently passing everything as an int32, whereas all those codes refer to hex. Which may be a problem, but I don't believe it is.

Okay thanks, I'll come back to that later then as those bool values are something I make quite a bit of use of in one of my apps

PCAssistSoftware:
> > Do you want to continue trying bits in this PR relating to appveyor before I merge it? quite happy to do so if you wish - no hurry on my part
> 
> We know the tests pass. I can play with my repo to see if it's worth doing a pr on the AppVeyor stuff.
> 
> If you are satisfied here, merge away.

Okay will do - thanks

PR #158, opened 2022-03-06
Provide Base Unit Tests for ProgressEstimator

RFBomb:
@PCAssistSoftware 

I was working on your branch when I poked around unit testing, and it would have been difficult to port it into my own repo, so I converted it over to a new branch on your repo (sorry!)

But this has 3 unit tests and is set up with a framework to compare the results of a run to the results of ProgressEstimator. 
Complete with 0-2kb files in the 'TestFiles' folder (which are copied to the BUILD folder when this is built, and all the tests are run from there and into `C:\\RoboSharpUnitTests`


PCAssistSoftware:
> @PCAssistSoftware
> 
> I was working on your branch when I poked around unit testing, and it would have been difficult to port it into my own repo, so I converted it over to a new branch on your repo (sorry!)

That's fine - no problem at all :)

> But this has 3 unit tests and is set up with a framework to compare the results of a run to the results of ProgressEstimator. Complete with 0-2kb files in the 'TestFiles' folder (which are copied to the BUILD folder when this is built, and all the tests are run from there and into `C:\\RoboSharpUnitTests`

Excellent work - will take a proper look in the morning when I have more time

RFBomb:
It should be pretty easy to set up all the tests you want to do. Do that in the 'ProgressEstimatorTests' file. 
I think you will just have to copy-paste the 'Sample Test Method' i put at the top of the file and set up all your various test conditions. 

Each method will run the test once, so you would want 1 method for each test condition ( ignoring files is already done, ignore larger is done).

PCAssistSoftware:
Just had a quick look, and that is brilliant, really love what you have done here.

Had always seen other projects which used Unit Testing but never had the time to look into it or use it myself, it is such a useful and powerful tool, really like how you can see all the tests and the results within Test Explorer.

Do you want to do any more work on it here in PR, or is it ready to merge in and then continue development / make any changes once it is in the main project?

RFBomb:
I want minor work done on it before main merge, tests for statistics objects methods should be done (VS can likely generate mockup file for it automatically). 

I'm leaving setting up the robocopy test cases to you since you have a lot more uses for option than my projects seem to. 

And I have an idea to test list only with no additional tests being written, but need to see if it will actually work.

RFBomb:
For your file locked test, use:

```
// files should already be in destination
Using ( var temp = file.OpenWrite(FilePath)
{
Run the test
}
// Assert
```

That using statement opens the file up for writing, which should cause that error you were recreating with your powershell script

I can write this one since I want to test the error generation anyway

PCAssistSoftware:
> I want minor work done on it before main merge, tests for statistics objects methods should be done (VS can likely generate mockup file for it automatically).
> 
> I'm leaving setting up the robocopy test cases to you since you have a lot more uses for option than my projects seem to.
> 
> And I have an idea to test list only with no additional tests being written, but need to see if it will actually work.

Yes that's fine with me, got a busy day of work today but then free tomorrow so will devote some time to it then

PCAssistSoftware:
> For your file locked test, use:
> 
> ```
> // files should already be in destination
> Using ( var temp = file.OpenWrite(FilePath)
> {
> Run the test
> }
> // Assert
> ```
> 
> That using statement opens the file up for writing, which should cause that error you were recreating with your powershell script
> 
> I can write this one since I want to test the error generation anyway

That sounds ideal for testing locked files / recreating errors

RFBomb:
ListOnly tests created - I'm proud of the simplicity here. I just created a derived class and overrode a bool. Happy it worked. 

FileInUse test - This was created and is currently failing, which is what the other PR aims to correct.

RFBomb:
We pushed at same time. I have conflicts I have to correct, hold off on more commits briefly please

RFBomb:
Alright, I'm done with this for now, Attribute Testing has been updated and is passing (except for Encrypted and Compressed, but those flags may not be forceable programmatically like the rest are). 
Feel free to merge this whenever you are satisfied I supposed. 

I know that the 'File In use' test is still failing, but I think that would be worthwhile to run alongside the other commit we have going on. 

Also, we probably want a test to ensure various events are being raised, such as OnError (since I seemingly broke it in the other PR!)

PCAssistSoftware:
All looks really good - and certainly going to prove useful.

I have been reading up on encrypted and compressed and there does seem to be issues setting them programmatically.

@tjscience is going to really like this i'm sure

As for the File In Use test I am confused as to why it is failing, as if I run the same job that the test is simulating in RoboSharp then it is working fine, and the results as far as I can see are as I would expect them to be in both ProgressEstimator and Results?

image


Unless anything else you want to change in light of my reply above then I am happy to merge it.

RFBomb:
Not sure,  it when I wrote the TestResult.Assert() method o evaluate all the statistics objects valleys to determine if any fail. 

If any do fail, it adds it to the 'errors' string list.

That list is output using Console.Writeline(), so if you view the output of the test you can see exactly why it's failing, as well as the robocopy log lines. 

I knew we'd want to view the log itself so I made sure it's always output into the test results, that way we can view it without having to use debugging mode.

RFBomb:
I KNOW WHY ITS FAILING!it just hit me. 

this is something I caught and fixed in the other PR already. The file is only marked as failed in this PR if OnError has an event listener. Since it doesn't, it's never marked as failed. 

So the test is working lol

PCAssistSoftware:
> I KNOW WHY ITS FAILING!it just hit me.
> 
> this is something I caught and fixed in the other PR already. The file is only marked as failed in this PR if OnError has an event listener. Since it doesn't, it's never marked as failed.
> 
> So the test is working lol

Ah, okay that make sense - I thought it was odd, as I was looking at all the output and it all looked okay, it was just expecting different values that it was getting

shall I merge?

RFBomb:
Yes good to merge, sorry for delay

PCAssistSoftware:
No problem at all, will do it now

PR #159, opened 2022-04-08
Create base interface for a RoboCommand Factory

RFBomb:
This PR adds a new object called RoboCommandFactory that contains methods for creating new robocommands. 

This object simply translates the constructors for `RoboCommand` into methods within an object. 
But the real reason for the PR is to introduce the `IRoboCommandFactoryBase` interface, which is meant to be a base interface for any objects that will be used to generate IRoboCommand objects. 

For example, in my dll that I am building, I require a specific set of constructors to generate my robocommands. These were custom-built constructors used for a class derived from RoboCommand. Since refactoring out of the main project into a dll, I now need to essentially allow the dll consumer to specify a factory object to create the robocommands. Only way to enforce the constructor requirements here is via an interface imposed on an object, then allow the consumer to design their own object that implements the interface. Since other consumers may one day need something similar, I introduced the factory here for their use.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ obj
 /packages
 /*.sln.GhostDoc.xml
 RoboSharp/RoboSharp.csproj
+nuget.config
--- a/RoboSharp/CopyOptions.cs
+++ b/RoboSharp/CopyOptions.cs
@@ -23,6 +23,19 @@ namespace RoboSharp
         public CopyOptions() { }
 
         /// <summary>
+        /// Create a new CopyOptions object with the provided settings
+        /// </summary>
+        /// <param name="source"><inheritdoc cref="Source" path="*"/></param>
+        /// <param name="destination"><inheritdoc cref="Destination" path="*"/></param>
+        /// <param name="flags"><inheritdoc cref="CopyActionFlags" path="*"/></param>
+        public CopyOptions(string source, string destination, CopyActionFlags flags) 
+        {
+            this.Source = source ?? string.Empty;
+            this.Destination = destination ?? string.Empty;
+            this.ApplyActionFlags(flags); 
+        }
+
+        /// <summary>
         /// Clone a CopyOptions Object
         /// </summary>
         /// <param name="copyOptions">CopyOptions object to clone</param>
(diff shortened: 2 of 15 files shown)


PCAssistSoftware:
Hi @RFBomb, assuming this doesn't break anything in existing project then happy to merge this for you when you are ready.  Not sure how to test it.

As you have incremeted version number might just be worth changing copyright year in RoboSharp.csproj to 2022 at same time as well

PCAssistSoftware:
@RFBomb - I assume the complete lack of response to my message means it is not ready to merge yet?

RFBomb:
@PCAssistSoftware 
- Sorry for delayed response (to your message that I can't seem to locate but have in an email from github?), this was just something that I whipped up while working on a dll that utilizes this dll. 

I will work on updating the unit tests for this to show how it works. But essentially its just a static object that contains two methods that merely point to the constructor for RoboCommand. 

The idea behind this is that if a consumer (like myself) wants to create a class that overrides the functionality of robocommand, and pass that factory to another consumer, that can now be facilitated. 

For example, I have a dll that will be creating robocopy commands to sync folders from one place to another. The consumer of that dll will be two seperate programs. One of them requires enabling logging, the other does not. Now, each of those 2 programs simply creates the factory object, passes it to the dll that generates the commands, and voila. The dll generated the commands according to the specifications of the factory methods defined by each consumer.

PCAssistSoftware:
Hi @RFBomb - that is odd, I had replied twice, but both messages gone for some reason??!

Let me know when ready to merge and I will get it done for you :)

RFBomb:
@PCAssistSoftware 
Check this last push out. 
This commit should hugely clarify what my design intent was for the factory. Essentially, a way for consumers to more easily create the robocommand (and any custom derived objects). 

So instead of using the constructors to produce the command, the factory can be called. The factory has been reworked from it (dumbly) just calling the constructor to now actually being an object to more easily create consumer factories.
 (Overriding the parameterless method to return a custom IRoboCommand instead of RoboCommand object affects the entire factory automatically). 

I have also introduced 2 new enums here. One resides within CopyOptions and the other within SelectionOptions for more compact construction of those objects (and thus eliminating potentially large amount of code thanks to also being able to use them via the factory methods)

PCAssistSoftware:
That all make sense - I like what you have done here, will be very useful - great work (as always)

RFBomb:
Ok, I believe this is ready for merge, unless you wanted to test it/have input on it. 

This is an example how I'm using it:
 - I have a 'RoboCopy' class that I created that is derived from RoboCommand. I override a few settings and enforce a couple defaults on it. 
 - I have a library I built that generates robocopy commands. It uses the IRoboCommandFactory interface to accept a factory object for generating the commands.
 - To enforce the library to build using the customized object, I pass the new factory object (shown in picture below) to the library. Then, all the commands it generated will be of my customized type, instead of the base robocommand type.
image

PCAssistSoftware:
All looks good to me, will go ahead and merge it now

PR #160, opened 2022-05-28
IRoboCommand & RoboQueue Fixes

RFBomb:
After the implementation of the factory method for RoboCommand objects, I realized that it required casting to work with the RoboQueue object. The casting was due to RoboQueue using RoboCommand and IRobocommand interchangably based on the writing process. Some methods called out RoboCommand, while the factory was producing IRoboCommands

This PR corrects that mistake to now use only IRoboCommand objects with RoboQueue.
This has the unfortunate side-effect of causing other objects to now also require passing out IRobocommand objects instead of RoboCommand objects, which arguably may be the correct thing to do since the events are also listed on the interface itself.

Update Version # to reflect significant changes ( Mainly that events are returning the  IRoboCommand objects instead of RoboCommand )

### DIFF
--- a/RoboSharp.BackupApp/MainWindow.xaml.cs
+++ b/RoboSharp.BackupApp/MainWindow.xaml.cs
@@ -547,7 +547,7 @@ namespace RoboSharp.BackupApp
         /// <summary>
         /// Disable associated MultiJob_CommandProgressIndicator to window
         /// </summary>
-        private void RoboQueue_OnCommandCompleted(RoboCommand sender, RoboCommandCompletedEventArgs e)
+        private void RoboQueue_OnCommandCompleted(IRoboCommand sender, RoboCommandCompletedEventArgs e)
         {
             UpdateCommandsRunningBox();
             Dispatcher.Invoke(() =>
@@ -586,7 +586,7 @@ namespace RoboSharp.BackupApp
         /// <summary>
         /// Log the Error to the Errors expander
         /// </summary>
-        private void RoboQueue_OnError(RoboCommand sender, ErrorEventArgs e)
+        private void RoboQueue_OnError(IRoboCommand sender, ErrorEventArgs e)
         {
             Dispatcher.BeginInvoke((Action)(() =>
             {
@@ -602,7 +602,7 @@ namespace RoboSharp.BackupApp
         /// Occurs while a command is starting prior to Robocopy starting (for example, due to missing source location), but won't break the entire RoboQueue. 
         /// That single job will just not start, all others will.
         /// </remarks>
-        private void RoboQueue_OnCommandError(RoboCommand sender, CommandErrorEventArgs e)
+        private void RoboQueue_OnCommandError(IRoboCommand sender, CommandErrorEventArgs e)
         {
             btn_StartQueue(null, null); // Stop Everything
 
--- a/RoboSharp.BackupApp/MultiJob_CommandProgressIndicator.xaml.cs
+++ b/RoboSharp.BackupApp/MultiJob_CommandProgressIndicator.xaml.cs
@@ -13,6 +13,7 @@ using System.Windows.Media.Imaging;
 using System.Windows.Navigation;
 using System.Windows.Shapes;
 using System.IO;
+using RoboSharp.Interfaces;
 
 namespace RoboSharp.BackupApp
 {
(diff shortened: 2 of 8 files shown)


PCAssistSoftware:
Hi Robert

Can you give me more details of what potential changes may have to be made to existing apps using RoboSharp this PR may cause (if any) - just want to make sure it doesn't break any of my (or others) existing apps (or at least if it does we have documented any changes needed) etc

RFBomb:
Essentially, RoboQueue had some methods accepting IRoboCommand and others accepting RoboCommand. This was an inconsistency that just wasn't caught prior to full release. That change will not affect any consumers, since RoboCommand implements the IRoboCommand interface. 

The change that may affect consumers is how the events are coded. 
 - Events originally were passing back `object sender`. At some point, we ( I ) changed this over to return RoboCommand objects instead. That change had zero effect on callers, since at the time all consumers were receiving objects, and casting into robocommand. This current change will now return the interface instead of the base robocommand object on all the events. 
 - Since the interface describes all of the functions of RoboCommand, the functionality available will not effectively change. But it will be made more clear that the type may not be the base RoboCommand object, but instead may be any object that utilizes the interface (such as a derived object)
 - The Signature of the methods will change though. This is what would break compiling for consumers.
    - This will only affect them if they relied on the `RoboCommand sender` in the signature. They would have to update it to be  `IRoboCommand sender` in their signature. 
    -  If they use implicit or `object` types in their event handler signature, this will not be a breaking change. 
        (implicit being something like an anonymous handler:   ` cmd.Completed += (o,e) => { // do something } `

RFBomb:
Essentially this is a consistency improvement. 

- RoboQueue is a list of IRoboCommand objects. 
- The factory produces IRoboCommand objects. 
- The interface describes events that should also return the interface as the sender, hence the change to the event signature. 


That said, this isn't ready for merge. As you say it failed on AppVeyor, so I want to double check it. My VS instance also complained about a compile error (then built successfully), so I want to double check that as well.

RFBomb:
Ok, reviewing the AppVeyor log reveals exactly what would break ( yay Unit Tests! )

The unit tests are what failed, since the event testing methods no longer matched up to the updated signature. On Tuesday I'll correct all this, double check my work, and come up with some sort of note detail describing the change.

PCAssistSoftware:
Okay, that makes sense - when it is ready I will test it with my existing apps as well so I can see what if any changes I need to make from a VB.NET point of view and then between us we will have covered all bases.

PCAssistSoftware:
Hi Robert

Just been testing this PR in my app to see if I needed to make any changes to my code to be able to compile and run - and I didn't!! - it worked perfectly fine without anything at all needing to be changed, I think mainly because in my app using VB I was declaring sender as "object" rather than 'RoboCommand' anyway so didn't need to change it to 'IRoboCommand' e.g

`Public Sub RoboCommand_OnCommandCompleted(ByVal sender As Object, ByVal e As RoboCommandCompletedEventArgs)`

Not sure why I had done it like that, and if that is "bad programming" or just an oversight but it worked before and still works now.

RFBomb:
The standard event signature returns object as the sender and 'eventargs' as the event occurrence. 

Anything that utilizes that signature can respond to any event, even ones it wasn't meant for. It's not bad practice, but you just may get an invalid cast if you are casting to object to something without checking type first. 

When the change was made to make the sender as RoboCommand, those older signAtures didn't break because it's still valid. It just clarified what type of object was being sent in the signature in order to eliminate need for casting. 

This change that makes it to 'IRoboCommand' will bring back the need for casting if an object of type 'RoboCommand' is required,  it since the interface exists it would be better for consumers to simply use the interface instead, since it provides all the same stuff. 

So like I said before, not a completely breaking change, and will only affect projects that use the signature after the first signature update.

RFBomb:
All that said, I believe this is good to merge if you think it is. It's passing all its compiles and tests (the tests didn't break, the UI did due to the signature change)

PCAssistSoftware:
> The standard event signature returns object as the sender and 'eventargs' as the event occurrence.
> 
> Anything that utilizes that signature can respond to any event, even ones it wasn't meant for. It's not bad practice, but you just may get an invalid cast if you are casting to object to something without checking type first.
> 
> When the change was made to make the sender as RoboCommand, those older signAtures didn't break because it's still valid. It just clarified what type of object was being sent in the signature in order to eliminate need for casting.
> 
> This change that makes it to 'IRoboCommand' will bring back the need for casting if an object of type 'RoboCommand' is required, it since the interface exists it would be better for consumers to simply use the interface instead, since it provides all the same stuff.
> 
> So like I said before, not a completely breaking change, and will only affect projects that use the signature after the first signature update.

That makes sense and is what I thought :)

PCAssistSoftware:
> All that said, I believe this is good to merge if you think it is. It's passing all its compiles and tests (the tests didn't break, the UI did due to the signature change)

Sounds good to me, will merge it for you now.

PR #167, opened 2023-02-23
Update Net4.0

tboehnlein:
I ran into an issue where I tried to copy from folders that had a single quote in the folder name.  Current code fails to copy such files.  With this change, the files copy without issue.

I am updating the 4.0 NET branch because this is what I have to use in my software to maintain compatibility with legacy computers.

### DIFF
--- a/RoboSharp/ExtensionMethods.cs
+++ b/RoboSharp/ExtensionMethods.cs
@@ -22,9 +22,8 @@ namespace RoboSharp
 
         public static string CleanDirectoryPath(this string path)
         {
-            // Get rid of single and double quotes
-            path = path?.Replace("\"", "");
-            path = path?.Replace("\'", "");
+            // Get rid of leading/trailing for single and double quotes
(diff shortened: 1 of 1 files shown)


PCAssistSoftware:
@Stocklone - apologies for the delay in getting round to looking at this properly and testing it.

@All

It is clear from testing that the existing method does cause problems if the path has a folder name which contains a single quote.

As you can see below using the example source path of "C:\tmp\Bob's Computer Barn" that the existing code actually removes the single quote from within the path (Bob's becomes Bobs) making the source path invalid so the copy obviously then fails.

 // Get rid of single and double quotes
path = path?.Replace("\"", "");
path = path?.Replace("\'", "");

image

Whereas the suggested fix does not get rid of the single quote in the middle of the path (Bob's) so the copy works fine.

// Get rid of leading/trailing for single and double quotes
path = path.Trim('\'', '\"');

image



Unless anyone objects I will merge this tomorrow

PR #168, opened 2023-02-23
Fixed issue with paths containing single quotes within

tboehnlein:
Paths with single quotes within the path do not copy correctly. 

Example: C:\temp\Bob's Computer Barn\myBarn.txt will not copy.

### DIFF
--- a/RoboSharp/ExtensionMethods.cs
+++ b/RoboSharp/ExtensionMethods.cs
@@ -65,9 +65,8 @@ namespace RoboSharp
 
         public static string CleanDirectoryPath(this string path)
         {
-            // Get rid of single and double quotes
-            path = path?.Replace("\"", "");
-            path = path?.Replace("\'", "");
+            // Get rid of leading/trailing for single and double quotes
+            path = path.Trim('\'', '\"');
 
             //Validate against null / empty strings. 
             if (string.IsNullOrWhiteSpace(path)) return string.Empty;


PCAssistSoftware:
@Stocklone 

Thanks for this - sorry for delay in responding - now looking into it.

@tjscience @RFBomb 

Hi Guys, hope you are well - can confirm that the issue reported here is definitely a problem and from my quick testing this evening that the proposed commit does indeed seem to resolve it.  Just wanted some comments / input from you if you had time in case you could see any issues or side effects before I potentially merge the change?

PCAssistSoftware:
@Stocklone - apologies for the delay in getting round to looking at this properly and testing it.

@All

It is clear from testing that the existing method does cause problems if the path has a folder name which contains a single quote.

As you can see below using the example source path of "C:\tmp\Bob's Computer Barn" that the existing code actually removes the single quote from within the path (Bob's becomes Bobs) making the source path invalid so the copy obviously then fails.

 // Get rid of single and double quotes
path = path?.Replace("\"", "");
path = path?.Replace("\'", "");

image

Whereas the suggested fix does not get rid of the single quote in the middle of the path (Bob's) so the copy works fine.

// Get rid of leading/trailing for single and double quotes
path = path.Trim('\'', '\"');

image



Unless anyone objects I will merge this tomorrow

Issue #177, opened 2023-09-19
Writing Unit tests


sophiedaiter:
Hi,
i have a method calling StartAsync which returns RoboCopyResults
i would like to mock startAsync in my unit tests and return different results (for happy flow and fail flow) but RoboCopyResult c'tor is internal. so i can't create an instance of it . 

what can i do? @RFBomb 
Thanks
Sophie

RFBomb:
Sophie,
I don't think any of us considered that scenario when we hid the constructor. It's also why I did a ton of work to the repo trying to support custom implementations of RoboCommand though. So I believe I can guide you in the correct direction here. 

Unfortunately, the work I did in #161 was never merged in, mostly because it's a major re-assessment of the library. But that fork might have what you need. It's what I use in my production project. 

@PCAssistSoftware - I've done a ton of work on #161 and outside it as well. I think it's likely a good time to have a discussion on evaluating it,  and either merging it, putting it out as a secondary package, etc. I'm personally happy with bringing it under my namespace in my GitHub if that's how we want to proceed. Doing it like that keeps the custom implementation logic isolated from this project, and will require just publicizing some of the constructors to accommodate it.

PCAssistSoftware:
Hi Robert, good to hear from you.

Happy to take a look at what you have done and discuss the way forward.  The second option may be the best way perhaps so that we don't impact too much on the existing package, but can decide that once discussed and evaluated, just need to be aware of the many people who use this library and any changes that may affect them, especially as any version number increment we may decide to do at some point will result in an automatic NuGet release etc.

RFBomb:
Yea, that reasoning is exactly why we never merged it in.
Most of the PR is meant to facilitate custom implementations of IRobocommand. I have created several myself, as I'm using it as the backend for a file sync framework I designed for my application (RFBCodeWorks.NetworkSyncObjects). 
I think it may make the most sense to have it just be a secondary nuget package, either contained within this repo, or even an external one. (My personal preference is a new project within this repo, that just builds the secondary package. This would keep it centralized to this repo instead of having to bounce between one the main one and one hosted on my page)

If I recall correctly, most of the additions are static helper methods to help with the evaluation/comparison of file pairs for custom implementations.

I could begin work on that sometime this week or next.

PCAssistSoftware:
That make sense to me and i'm happy to test and evaluate as needed, and we can then decide location as best to see what Terry has to say if around as well.

RFBomb:
[...]

PR #161, opened 2022-06-28
RoboSharp.Extensions Namespace to provide for custom IRoboCommand Implementations

RFBomb:
Various issues fixed for WPF .net5 compatibility

image

PR #179, opened 2023-09-26
Minor Fixes and Validation

RFBomb:
This PR is a subset of the changes proposed in #161.
I've contained it to what should be the safest changes to implement on the core library.

Some of these changes were designed to facilitate custom IRoboCommand implementations, so I left those in.
The major changes to this are primarily publicizing / adding constructors to various objects to allow for either external unit testing or custom implementations.

I also have a replacement for the ObservableList<T>, which is used by RoboQueue and RoboCopyResultsList.
Now, I'm not saying I regret RoboQueue, because I learned A LOT doing it, but I do think that writing ObservableList was an err on my part. I wound up writing a 'ConcurrentList' which has immutable collections to play nicer with UI updates when multiple results are updated at same time. This is NOT part of this change list, as I think it should be isolated for review.

Linking #177  as it will be resolved in this PR.

The only potentially breaking change here was moving the enums out of SelectionOptions and CopyOptions into the default namespace, which makes it overall easier to use long term. (My bad for putting it in there initially)


RFBomb:
Rebased after #180

RFBomb:
I want to say I believe this is mostly finalized. I'm starting work on the Extensions namespace shortly, which will include a custom implementation I created to more efficiently MOVE files that reside on the same root drive. ( using File.Move() is significantly quicker than robocopy if on same drive)

PCAssistSoftware:
All appears okay to me, happy to squash and merge if/when you are ready?

RFBomb:
@PCAssistSoftware Thanks for testing all that so quick!
I didn't have internet for a bit and got some more relatively small changes put in.
- Some unit testing for CopyOptions and SelectionOptions methods.
- Minor XML Doc fixes, cleaning up some more compiler warnings/messages.

BIG CHANGE: commit ( https://github.com/tjscience/RoboSharp/pull/179/commits/acbddeb9910bc42c5a277a523260dd7d4a6d6cc9 ) needs some verification on your part, as I know you actually use this functionality and I have nothing to test against. 
Basically, that reworks the authentication to expose it to custom implementations. The actual functionality should be unchanged, but its been moved to a new 'Authentication' static class.  ( This also allows to potentially to authenticate source and destination separate from each other in custom commands ).

I know it didn't pass the current check, I'm investigating that now.

PCAssistSoftware:
@RFBomb - no problem at all, having a quiet day workwise so thought I would devote some time to this.

re big change / ( [acbddeb ](https://github.com/tjscience/RoboSharp/commit/acbddeb9910bc42c5a277a523260dd7d4a6d6cc9)) - actually I don't use the authentication, I recently merged in some changes by @gabrielgt relating to authentication support for .NET 6.0 and had to specifically setup a test machine in a domain to make sure that worked, but that was first time I had ever even passed a username and password to RoboCommand StartAsync.   I can resurrect that test environment and test again with this PR for you.

PCAssistSoftware:
Can't build current PR, got error on line 102 of authentication.cs - user of unassigned local variable 'result'  - that might be why it wasn't passing current check?

RFBomb:
So, the check fails were also present on my computer, because I couldn't build the unit test framework for some reason (even before I added a test). Not really sure what that was about, but I'm updating the project file for the unit tests to the modern Project-SDK style, which resolves it. 

As far as that line 102, I'm really confused by that, as line 102 is the return statement. I cannot build on Net6.0 on my machine (stuck on VS2019), but I believe I know the issue and will have a fix up mometarily. Still investigating the unit tests failing though.

PCAssistSoftware:
Have tested authentication as much as I can by passing domain, username and password to RoboCommand StartAsync and it appears to still work as before.  Passing invalid credentials fails as expected with Win32Exception of 'The user name or password is incorrect' being thrown in ImpersonatedUser.cs and passing correct credentials results in filecopy completing successfully, so as far as I can see all okay in relation to that point.

RFBomb:
Thanks for testing all that! I was confident it would still work, but needed it tested with the updated static class. Consumers should see no difference in it. 

I have just resolved all the issues the unit new tests brought to light. in this latest commit as well. ( I'm really glad those other unit tests were written to validate everything before, as its really brought my confidence level up in these changes)

PCAssistSoftware:
no problem at all - would you like me to merge it now?

RFBomb:
If you are happy with it, I'm happy with it. 
I'm fairly confident this covers all the bases I needed to provide the Extensions package its functionality.
I'm excited to bring over my 'RoboMover' object I created as well, as its significantly quicker than RoboCopy when performing a MOVE action on the same drive ( ex:    F:\Source -> F:\1\2\3\Destination )

PCAssistSoftware:
consider it done :)

looking forward to seeing the Extensions package

PR #180, opened 2023-09-26
Rework ObservableList for improved thread-safety

### LINKED ISSUE #165: ObservableList<T> and WPF - Not Truly Safely Observable
I'm writing this into an issue instead of just fixing it since I don't want to include unnecessary components into the library unless its truly warranted (funny right? considering all my other changes).

-----
### The Problem:

Basically, the ObservableList<T> object was written to have a thread-safe list to enable the functionality of RoboQueue and ProgressEstimator. Without it, those two classes won't work at all. 

The problem is for consumers that attempt to use it alongside WPF views. For infrequent updates, this object works fantastic. What I mean by 'infrequent updates' would be something like generating a popup, having the user select from the list, then those items get added to another list. Since all of those events are generated from the UI thread, and they also take time to resolve the opening/closing of the windows, the view is updated just fine. 

The problem occurs when attempting to modify the list faster than the UI updates the list. You may get an exception such as:
```
System.InvalidOperationException: 'A collection Add event refers to item that does not belong to collection.'
```
When multiple events are dispatched to the UI, the UI handles them whenever it handles them. If the modification occurs while the UI is updating, you can get a fault due to the Count property (as read by the UI) being 6 when the UI starts displaying the items, but by the time the UI finishes iterating over the list there are 7 items.

An example of this would be if you had an `ObservableList<string>`  that was holding the log lines in a custom implementation. As the custom implementation processes the files/folders, it will update this list, but it will likely update much quicker than the UI handles the `CollectionChanged` events. ( This is actually the exact scenario I ran into this issue - I was reacting to the ProcessedFile event in the ViewModel, and adding those to a list to update the View)

-----
### The Solution:

I've done some googling, and what I found was this article, and the corresponding library:

https://www.meziantou.net/thread-safe-observable-collection-in-dotnet.htm

https://github.com/meziantou/Meziantou.Framework/blob/main/src/Meziantou.Framework.WPF/Collections/
 
Implementing a design like above will require the `<UseWPF>true</UseWPF>` declaration in the project. 
I did not want to add that requirement to this library without consensus of consumers and other maintainers. 
Adding this will effectively add WPF requirement to any consumers that are not using the NetStandard version of this library, even if their project isn't using WPF (ex: they are doing a winforms or console project).

----
### Potential Resolutions:

This library only uses this object privately, and is expsoing it via the ReadOnlyCollection object.

We could make it internal-only, therefore removing the issue from consumers by simply not allowing them to instantiate the object. This of course will break any implementations by consumers that rely on this object since it has been introduced. But it would force them to use an actual WPF-Safe object. 

We could rename it to `ConcurrentList<T>`, then add a new project to the solution that provides the `ObservableList<T>` object for consumers in the RoboSharp namespace, using a similar implementation to the link above. This will only break consumers if they choose not to include that project reference. This also adds complexity to not only this library, but to consumers. As such, I don't think this is a great route when a library like the linked one above exists.

As of writing this, I say we leave it exposed, but rename it `ConcurrentList` to more accurately describe its use case. `ObservableList<T>` will be marked as Obsolete, and throw an exception if you try to instantiate it, describing why it threw. This will effectively alert consumers (by breaking their uses of it) that it is NOT UI-Safe. This change will allow it to still be utilized by custom implementations, but will require those projects be updated to the new object name.

RFBomb:
Resolve #165

- Marks ObservantList<T> as obsolete.
- Introduces ImmutableList<T> ( Internal only ) and ConcurrentList<T> (public) as the replacement. 

These two collections are used by the RoboQueue and RoboCopyResultsList objects. These new implementations are much more thread-safe, and should play nicer with UI should they be used.

Note: Proper UI binding is on ConcurrentList.ReadOnlyList property, as per the remarks on the class.


RFBomb:
The AppVeyor build failed earlier due to the 'Obsolete' tagging of the class, so it was working as intended. 
But that made me think hard about it, and the core purpose of the class was staying the same. 
As such, this latest commit removes the 'ConcurrentList' and merges those changes into ObservableList. 

The niche-case methods [ such as Sort(bool verbose) ] have been removed. Those were originally meant to work around WPF compatibility, but is now handled by the event sender. ( If the consumer resides in the System.Windows.Data namespace, it sends Reset instead of Add/Remove/Etc)


When this is merged in, I'd recommend using 'Squash and Merge' to avoid inserting then removing the ConcurrentList<T> file into the history of the main branch.

RFBomb:
@PCAssistSoftware 
I believe I'm quite satisfied with this PR, so review at your leisure. 

I'm still working on the other PR, though it's more double-checking that one at this point. The Extensions namespace will rely on both of these, PRs, so once they are merged in I'll set that up, that way it's a cleaner PR for extensions.

PCAssistSoftware:
@RFBomb  

Okay cool, will test in next couple of days and if all okay will get it merged in for you

PCAssistSoftware:
Hi @RFBomb - all looking okay to me - do you want me to get this one "squashed and merged" now for you or wait for the other one?

RFBomb:
This one is able to commit on its own. 

My goal for the other one is to review it today (by putting in all my extension stuff and just keeping those items local) to verify nothing else is needed. I think I have it all right now, but ill leave a comment on that after I do the above check.

PCAssistSoftware:
okay will do it now then

PR #182, opened 2023-10-13
Fix CopyOptions.Compress Test

RFBomb:
I realized last night that the test I created to allow consumers to test if they can utilize /COMPRESS was flawed. It obeyed the safeguard instead of testing it. Resulting in potential false positives.

The new test bypasses the safeguard, as was initially intended.
This also increments the package version as to push out the updated test logic. 

I also verified that the ProgressEstimator was working properly with multithreaded settings, which I know used to not be the case. I guess all the thread-safety improvements I've done resolved that, but its now validated within a test.

Also, while testing this I was running into the Encoder(437) issues yet again. So I reworked that logic, primarily within the RoboSharpConfiguration, to ensure that the encoding is initialized prior to attempting to create the configurations that utilize it. ( Prior to this, the encodings were only initialized if creating a LoggingOptions object prior to a RoboSharpConfiguration object. That nuance is resolved here )

### DIFF
--- a/RoboSharp/ApplicationConstants.cs
+++ b/RoboSharp/ApplicationConstants.cs
@@ -1,9 +1,25 @@
 ﻿using System.Collections.Generic;
+using System.Text;
 
 namespace RoboSharp
 {
-    internal class ApplicationConstants
+    internal static class ApplicationConstants
     {
+
+        /// <summary>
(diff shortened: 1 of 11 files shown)


RFBomb:
(This is ready for merge, small fixes on this one)

PR #183, opened 2023-10-15
v1.3.3 - Validate & Fix JobFiles, add Serialization helpers

RFBomb:
@PCAssistSoftware 
Try this PR and let me know if you have any issues on it. 
- This includes a new static method for RoboCommand 'LoadFromJobFile'
- This also includes some fixes and unit tests. Also introduced is a method for Authentication of the JobFile save path. 

I know there is a case where RoboCopy will fail to create the file. I'm not sure how to write a unit test for that to force the failure though. I'm also unsure how we should modify it for further evaluation. What I mean is that currently the output of RoboCopy itself is being effectively ignored. The events for RoboCommandError will still be raised, but other events won't be (they shouldn't be raised because it shouldn't actually be processing any files during this save). Technically, RoboCopy will spit out results, so we should be able to return them to callers after a save is completed. But at original time of writing, I didn't consider that.

Also, take a look at JobOptions. I had committed a 'LoadJobFilePath' and a 'FilePath' properties. Should 'FilePath' be renamed to 'SaveFilePath' for clarity? ( Note that doing so will break anything that uses it until consumers rename it, so this is likely ill-advised)


RFBomb:
After some testing, I noted the following:

Job Options 
 - NoDestinationDirectory
 - NoSourceDirectory

The above properties are not actually switches RoboCopy reacts to when producing a job file. Instead, they are flags that get written into the job file itself to indicate to robocopy's parser that no location was specified.

As such, I think that the Save functionality should be reworked, or the JobOptions object should be reworked. Either set it up so that those are Get-Only (they perform a check against the CopyOptions, similar to how robocopy works when writing the file). Or we update the Save functionality to remove the two boolean parameters that are used during saving. Personally, I think making these Get-Only makes more sense, as it more accurately reflects how robocopy functions. Then keep the parameters in the method, since they can be used to explicitly declare it won't be saved to the file.

The current commit introduces this change.

PCAssistSoftware:
Thank you - I can confirm with this commit that I no longer get the System.NullReferenceException in RoboCommand.cs when using SaveAsJobFile, and also that when using below the values are populated correctly.

 ```
Dim x As JobFile = JobFile.ParseJobFile("C:\Users\darre\OneDrive\Desktop\job.rcj")
q.AddCommand(x)
```

**Comment by PCAssistSoftware** [comment]
> After some testing, I noted the following:
> 
> Job Options
> 
> * NoDestinationDirectory
> * NoSourceDirectory
> 
> The above properties are not actually switches RoboCopy reacts to when producing a job file. Instead, they are flags that get written into the job file itself to indicate to robocopy's parser that no location was specified.
> 
> As such, I think that the Save functionality should be reworked, or the JobOptions object should be reworked. Either set it up so that those are Get-Only (they perform a check against the CopyOptions, similar to how robocopy works when writing the file). Or we update the Save functionality to remove the two boolean parameters that are used during saving. Personally, I think making these Get-Only makes more sense, as it more accurately reflects how robocopy functions. Then keep the parameters in the method, since they can be used to explicitly declare it won't be saved to the file.
> 
> The current commit introduces this change.

I also agree that Get-Only makes more sense for both of those options

**Comment by PCAssistSoftware** [comment]
> Should 'FilePath' be renamed to 'SaveFilePath' for clarity? ( Note that doing so will break anything that uses it until consumers rename it, so this is likely ill-advised)

I think it is okay as it is, the documentation explains its purpose well enough so not worth the hassle of changing it, especially as it will break for existing users etc

**Comment by RFBomb** [comment]
>I also agree that Get-Only makes more sense for both of those options

I played with it for a moment. I realized that we would actually need to pass either the COpyOptions or the RoboCommand object itself into the constructor for JobOptions to get the 'Get-Only' functionality. As such, removed those two properties, since they had no effect on the actual output of robocopy, as verified by several tests. 

Instead, those are now housed within the JobFile object, since that is meant to be an indication of what was parsed out from the file.

**Comment by PCAssistSoftware** [comment]
Thanks for looking at this today - much appreciate it! All seems to be working okay now from what I have tested.  Let me know if anything else you want to check or modify, if not then we can merge these changes in later today or tomorrow perhaps.

**Comment by RFBomb** [comment]
I don't use this functionality, so if you think that its good I'm good with it. 
~~I did not increment the version # though, i can do that if you think its ready~~

Version incremented.

**Comment by PCAssistSoftware** [comment]
> I don't use this functionality, so if you think that its good I'm good with it. ~I did not increment the version # though, i can do that if you think its ready~
> 
> Version incremented.

Thanks, will test again tomorrow and if I don't find any issues then I will get it merged :)

**Comment by PCAssistSoftware** [comment]
Question.

In RoboCommand we declare the variable for the job name / ID Tag as "Name" and I use this a lot in my app, for example when listing all the jobs in a grid etc and it works fine.

I have now noticed that in JobFile.cs its name then changes to Job_Name which then breaks my code when loading a jobs from a .rcj file as above, because all the code is looking for Name

For example in the RoboQueue_CollectionChanged event I check e.NewItems.Item(0).Name but when adding a job from a .rcj job file this fails because "MissingMemberException: Public member 'Name' on type 'JobFile' not found.", and after drilling down it is because it is seeing Job_Name instead at this point.

So my question is - any reason it is called Name everywhere else, but Job_Name here? and any reason it can't be changed to be consistent so it is Name here as well?   thoughts please?

**Comment by PCAssistSoftware** [comment]
Not necessarily related to jobs, but as we have been talking on this issue I will ask it here:-

In previous releases of RoboSharp when just adding the basics of a source, destination and setting copy subdirectories including empty if I looked at my command options it would look like below:-

image

Something has changed in a recent commit as now the same settings is adding /BYTES a second time and also now adding /TEE which was never added before?  any clues?

image

This changed somewhere between 1.2.8 and 1.3.0

**Comment by RFBomb** [comment]
> For example in the RoboQueue_CollectionChanged event I check e.NewItems.Item(0).Name but when adding a job from a .rcj job file this fails because "MissingMemberException: Public member 'Name' on type 'JobFile' not found.", and after drilling down it is because it is seeing Job_Name instead at this point.
So my question is - any reason it is called Name everywhere else, but Job_Name here? and any reason it can't be changed to be consistent so it is Name here as well? thoughts please?

Probably an oversight. No reason why it can't match. 
That being said, I'm confused as to why you would get that exception. RoboQueue is a collection of IRoboCommands, and JobFile implements that interface. As such, it should be retrieving the Name from the interface, not the object properties.


> Something has changed in a recent commit as now the same settings is adding /BYTES a second time and also now adding /TEE which was never added before? any clues?

Bytes is forced starting at this commit: https://github.com/tjscience/RoboSharp/commit/b8352424a8e223b8eca94fdaef7aa71ed39fa1d5

It has always been part of the LoggingOptions, so not sure what else would have changed. I did some searching and didn't notice anything else. that would print it. But that appears as if its turned on in logging options (first instance) then getting added forcefully (second instance).

Which makes sense, given my change here (which I didn't think would cause an issue), as it was turning on the options required for RoboSharp to get data. 

https://github.com/tjscience/RoboSharp/blame/582e8e71a4660082e47b78d737668ae7e9d61751/RoboSharp/LoggingOptions.cs#L21

- I didn't realize that /BYTES would be set twice. 
- Verbose Output was determined required for processing EXTRA files in the Estimator, if I recall correctly.  
- OutputToRoboSharpAndLog - When specifying a log file path, robocopy outputs to that instead of the console std-out. /TEE is required to write to both locations. Without it, robocopy will only write to the console if a log file path has not been specified.

**Comment by PCAssistSoftware** [comment]
Thanks Robert, much appreciated - last commits have resolved /BYTES appearing twice and my Job_Name / Name issue :)

Still not sure why /TEE is now appearing when it wasn't before - as I am not specifying a log file path or doing anything differently than I was before.  If it is needed then not a problem, really not an issue it being required - so just really trying to understand its appearance now in later releases when it wasn't before in case it may affects anything else in my app

**Comment by RFBomb** [comment]
It was the constructor change, setting defaults required for certain library functions on construction. 

I could introduce a secondary evaluation for /TEE if you like. 
But as it is, It shouldn't have any effect if no log files paths were specified. 

image

If no logs are specified, I could omit the /TEE switch. Would be an easy couple lines if you like.

**Comment by PCAssistSoftware** [comment]
> If no logs are specified, I could omit the /TEE switch. Would be an easy couple lines if you like.

Think that would be nicer if not too much work - as think it looks nicer if command options only shows switches you have actually chosen to use - but only do it if it isn't going to create a lot of extra work

**Comment by PCAssistSoftware** [comment]
Minor issue with job file builder.

I am creating a robocommand with just source, destination and copy subdirectories including empty (/e) selected, and then saving the job file. 

The resulting .rcj file is perfect except it has BOTH copy subdirectories including empty (/e) AND copy subdirectories (/s) when it my case it should only be /e, and I believe should only ever be one or the other? as one **includes** empty and one **excludes** empty so would override each other if both supplied

**Comment by RFBomb** [comment]
So little misnomer here I think, but I'll take a look. 

The JobFileBuilder doesn't create the RCJ file, robocopy does. The JobFileBuilder parses the RCJ file into a RoboCommand, which is then used in the constructor for JobFile

**Comment by PCAssistSoftware** [comment]
> So little misnomer here I think, but I'll take a look.
> 
> The JobFileBuilder doesn't create the RCJ file, robocopy does. The JobFileBuilder parses the RCJ file into a RoboCommand, which is then used in the constructor for JobFile

Oh okay, well definitely when I put in a breakpoint and examine the robocommand after creating it, it only has /e in the command options at this point.  It is only the job file when saved which then contains both /e and /s

**Comment by RFBomb** [comment]
>  It is only the job file when saved which then contains both /e and /s

I wrote a unit test that is isolated from RoboSharp to produce the results. What you describe is simply the output of RoboCopy. /S is contained within /E, which is why both show up. 

image


image

**Comment by PCAssistSoftware** [comment]
> > It is only the job file when saved which then contains both /e and /s
> 
> I wrote a unit test that is isolated from RoboSharp to produce the results. What you describe is simply the output of RoboCopy. /S is contained within /E, which is why both show up.
> 
> image
> 
> image

Ah okay, thanks for confirming.  I came to the same conclusion when saving a job file directly from RoboCopy command line as well.  Not an issue now I know, just seemed counterintuitive when one is described as "including empty" and the other "excluding empty" so you would think it would save one or the other not both.

**Comment by RFBomb** [comment]
You might find this of use. I just created two new interfaces in the RoboSharp.Interfaces namespace, as well as a new object 'JobFileSerializer'. 
This iterates across a collection of IRoboCommand objects, saving to a specified directory. You can also load them back in using that object as well. ( It returns a new object that houses the enumerator which will read the files into JobFile objects)

**Comment by RFBomb** [comment]
Regarding #153 
This last commit introduced a RoboCommandXmlSerializer to the Extensions package.
Let me know what you think after playing with it. 

We could move this into the main package if desired, but it uses System.Reflection and System.Xml.Linq to perform the functions, so I wasn't sure if it should be in main package.

**Comment by PCAssistSoftware** [comment]
Interesting ;)

Assuming I am just passing a roboqueue to it like below then if so I am getting an error

q is my RoboQueue and contains two RoboCommand's which I have been testing with

```
Dim x As New RoboCommandXmlSerializer
x.Serialize(q, "C:\tmp")
```

image

**Comment by PCAssistSoftware** [comment]
Deleted last message as realized my mistake, was just passing a path to a folder (doh!), rather than a path to an xml file! - working now - will test further and report back....

**Comment by PCAssistSoftware** [comment]
Okay, first impression are - I like this!!

Serialize I have working and output looks fine.  

So how do I then use Deserialize to read the file and get the jobs back in a RoboQueue and ready to run?

**Comment by RFBomb** [comment]
Ok, this is where I wanted some feedback from you. 

Should the IRoboQueueSerializer interface's Deserialize method return the IEnumerable<IRObocOmmand> directly, or should it return an object that implements the Deserializer interface (which has a single method currently that returns the ienumerable)

**Comment by PCAssistSoftware** [comment]
> Ok, this is where I wanted some feedback from you.
> 
> Should the IRoboQueueSerializer interface's Deserialize method return the IEnumerable directly, or should it return an object that implements the Deserializer interface (which has a single method currently that returns the ienumerable)

Well personally how I would see it working is something like below potentially - but not sure if it helps answer the question or not?? or whether that is the best way to do it or not, but I certainly like the way this is heading ;)

```
Dim x As New RoboCommandXmlSerializer

x.Deserialize("c:\tmp\file.xml")

Dim q As New RoboQueue

q.AddCommand(x)
```

**Comment by RFBomb** [comment]
```
Dim x As New RoboCommandXmlSerializer

var collection = x.Deserialize("c:\tmp\file.xml")

Dim q As New RoboQueue

q.AddCommand(collection)
```

Updated.

**Comment by RFBomb** [comment]
~~I just took a peek at the Dependencies, and no change. so we should be able to include the XML Serializer implementation into the core if you prefer.~~

I think it makes sense to provide it in the main package, so I moved it there. It doesn't involve any extensions code, and synergizes with RoboQueue.

**Comment by PCAssistSoftware** [comment]
> I just took a peek at the Dependencies, and no change. so we should be able to include the XML Serializer implementation into the core if you prefer.

Fantastic work - thank you so much for doing this, is exactly what I needed and so much cleaner than creating multiple rcj files when you have a queue containing lots of (robo)commands.

~~I think it should stay in the Extensions package personally as it is all of your "hard" work so probably nicer for it to stay together in that one package which then clearly shows as you being the Author of it all - what do you think?~~

Will test more tomorrow to make sure nothing else needs doing and then if we are both happy we can get it merged in :) - thanks again for all your work on this

**Comment by RFBomb** [comment]
I was debating in having it stay in Extensions for the same Reason you stated lol. But I think it synergies too well with its primary use case, RoboQueue, to have it be an add on. 

I won't be available to work on it tomorrow, so let me know. I think I'm quite satisfied with it. I did increment the main package #, I dont think I incremented extensions # though (Which it will need). 

I also want to try to improve upon the XML serializer, as using reflection is sllooowwww, and the current implementation is calling it up fresh for every single RoboCommand that is getting parsed.

**Comment by RFBomb** [comment]
TLDR: Its good as is.

Ok, so I did some minor benchmark testing (didn't write a flesh out benchmark, just ran the unit test with various lengths of the array).

I was able to optimze the reflection pattern I used to serializing to XML. BUT the improvement is only noticeable when performing a serialization/deserialization of 50 commands or greater. Below 50 commands the current method beats it out by approx 10ms, which includes writing to disk.  50 is the approximate break-even point (consistent 4ms difference on my system). At 100 commands the 'optimization' beats out by approx 15ms. and at 400 commands it beats it by roughly 200ms. But realistically, I doubt anyone will be performing serialization of 50+ commands to a single XML file. But should that Serializer be kept in memory and used 50+ times over the life of the program, then it is a marginal improvement. ( I didn't test memory usage though)

For clarity, here was my approximate results on my system ( 5900X )
|  QTY |   3    |     50         |       100      |    400  |
| --- | --- | --- | ---| --- |
|  Without Optimization    |     130ms    |   160ms     |    200ms   |    480ms  |
|  With Optimization         |     152ms    |   156ms     |     180ms  |    256ms  |

So I'll keep the code I wrote, but I think the slightly simpler stuff currently uploaded is fine to keep.

Just for clarity here, here is the optimizer class (for future reference)

```
private class ReflectionOptimizer<T>
{
    public ReflectionOptimizer()
    {
        var props = typeof(T).GetProperties(PropertyBindingFlags);
        PropertyCount = props.Length;
        BooleanProperties = props.Where(p => p.PropertyType == typeof(bool)).Select(p => new PropertyOptimizer<bool>(p)).ToArray();
        StringProperties = props.Where(p => p.PropertyType == typeof(string)).Select(p => new PropertyOptimizer<string>(p)).ToArray();
        IntProperties = props.Where(p => p.PropertyType == typeof(int)).Select(p => new PropertyOptimizer<int>(p)).ToArray();
        LongProperties = props.Where(p => p.PropertyType == typeof(long)).Select(p => new PropertyOptimizer<long>(p)).ToArray();
        StringCollectionProperties = props.Where(p => p.PropertyType == typeof(IEnumerable<string>)).Select(p => new PropertyOptimizer<IEnumerable<string>>(p)).ToArray();
        StringLists = props.Where(p => p.PropertyType == typeof(List<string>)).ToArray();
    }

    public int PropertyCount { get; }
    public IEnumerable<PropertyOptimizer<bool>> BooleanProperties { get; }
    public IEnumerable<PropertyOptimizer<string>> StringProperties { get; }
    public IEnumerable<PropertyOptimizer<int>> IntProperties { get; }
    public IEnumerable<PropertyOptimizer<long>> LongProperties { get; }
    public IEnumerable<PropertyOptimizer<IEnumerable<string>>> StringCollectionProperties { get; }
    public IEnumerable<PropertyInfo> StringLists { get; }


    public class PropertyOptimizer<PropType>
    {
        delegate void SetterDelegate(T obj, PropType value);
        delegate PropType GetterDelegate(T obj);

        public PropertyOptimizer(PropertyInfo prop)
        {
            propertyInfo = prop;
            IsObsolete = prop.CustomAttributes.Any(attr => attr.AttributeType == typeof(ObsoleteAttribute));
            if (IsObsolete) return;
            bool locatedDefaultValue = TryGetDefaultValue<PropType>(prop, out PropType val);
            DefaultValue = locatedDefaultValue ? val : default;
            // GetMethod
            getMethod = (GetterDelegate)prop.GetGetMethod().CreateDelegate(typeof(GetterDelegate));
            setMethod = (SetterDelegate)prop.GetSetMethod().CreateDelegate(typeof(SetterDelegate));
        }

        public readonly PropertyInfo propertyInfo;
        public readonly bool IsObsolete;
        private readonly bool locatedDefaultValue;
        private readonly PropType DefaultValue;
        private readonly GetterDelegate getMethod;
        private readonly SetterDelegate setMethod;

        public bool IsDefaultValue(PropType value)
        {
            if (!locatedDefaultValue) return false;
            if (value is string str) return str.Equals(DefaultValue.ToString(), StringComparison.InvariantCultureIgnoreCase);
            return value.Equals(DefaultValue);
        }

        public void SetValue(T obj, PropType value) => setMethod.Invoke(obj, value);
        public PropType GetValue(T obj) => getMethod.Invoke(obj);
    }
}

```

**Comment by RFBomb** [comment]
For flexibility reasons, the XML serializer has an virtual method for generating the command. 

Would it make more sense to introduce a constructor that accepts a function instead? (Or even an IRoboCommandFactory)That way you wouldn't need to create a whole class to override which type of IRobocommand is built. It would use either the supplied function/factory, and if one isn't supplied use the default. 

My gut says simplicity go for a function, but since the IRoboCommand factory exists, it could be justified to pass it that reference. (realistically, both are easy to implement, so there is choice of both). 

I'm only thinking about this for the possibility of creating RoboMover and other custom implementations from instead of creating those from a RoboCommand. 

I do want to add a few more virtual methods for it before releasing. (My project for example has a custom LoggingOptions class)

**Comment by PCAssistSoftware** [comment]
Give me a bit more detail how each option would change (if it does change) how I would load from XML and then add to a queue and then I can give a more considered answer specifically in relation to how it might affect how I currently work, or how it might affect existing apps (again if it even changes them at all that is).   Have done a quick implementation this morning of the RoboCommandXmlSerializer with Serialize and Deserialize in my apps and adding it to queue using way you said [here](https://github.com/tjscience/RoboSharp/pull/183#issuecomment-1765195006) and it works like a dream

**Comment by RFBomb** [comment]
So essentially, RoboMover was created out of a desire to improve a specific use case of my project (moving files from one to another folder over UsB2.0). I also wanted to avoid writing the log file during ListOnly operation, as I'm using it for progress bars only. 

My NetworkSyncObjects library is effectively a complex RoboCommandFactory. Hence creation of that interface and base object in this library. 

So basically, when I create one, I do so via my factory. It provides a custom RoboCommand and custom LoggingOptions, as well as some default ignores. The factory evaluates it and decides to either create my Roboacommand object, or a RoboMover object. 


The proposal here is to allow the serializer to work with directly produce commands, such as RoboMover, if desired, via a factory method that evaluates the parsed options. ( default functionality would remain as is)

**Comment by PCAssistSoftware** [comment]
Okay that makes sense - appreciate the explanation.

**Comment by RFBomb** [comment]
Yea Basically, I have a very custom use case of what I'm doing with robocopy, and I've also had quite a bit of trouble with handling XML files that interact with a third-party application. As such, I was looking to make this XML serializer as flexible as possible, while still having relatively simple and functional, default application, as I don't anticipate the vast majority of people using the library to customize it as much as I needed to.

(I recently wrote a custom XML writer because the 3rd party required 'utf8' in the header to be capitalized, and the standard one provided forced it lower case. So that was a nightmare to figure out and fix why it was getting silently rejected)

**Comment by PCAssistSoftware** [comment]
Not had a huge amount to do with XML, but I know from reading other posts it can be a real pain sometimes so appreciate the frustration and work going into it.

**Comment by RFBomb** [comment]
Ok, I've re-written the XMLSerializer to provide much better customization support. If needed, consumers can override various methods to provide their own conversion into/out of XML for each Options object, and for the command itself.

Ready. To. Merge.

**Comment by PCAssistSoftware** [comment]
I have been testing and before I merge I do have an issue relating to removing jobs from the queue - it works fine (as it always has) with jobs added normally, but is throwing errors when removing jobs which have been added by way of the serializer.

In my app jobs are shown in a list view, and clicking a remove all button to clear the queue is fine.

But if I select a single item in the list and click to remove it then it throws an exception.  my remove button is simply using queue.removecommand and parsing the index of the item to remove.

The error occurs in the below event - wanted to mention it before I merge just in case it can be resolved.

Your thoughts would be appreciated 

image

**Comment by RFBomb** [comment]
To be clear, you are adding results to a List object as they complete. Then, you can remove the command from the RoboQueue, and when that occurs you also remove its results from the list ?

Writing a new unit test now to validate that.

**Comment by PCAssistSoftware** [comment]
> To be clear, you are adding results to a List object as they complete. Then, you can remove the command from the RoboQueue, and when that occurs you also remove its results from the list ?
> 
> Writing a new unit test now to validate that.

Actually I think this issue might be more specific to my app that I first thought, now trying to work out why.....   yes I have a bindinglist storing my results data for showing in a grid, when adding jobs normally the jobs each get added to this binding list, but for some reason when adding in bulk via serializer only one job gets added to by binding list, hence the error,  think I have missed something somewhere and the integration of your solution may not be as seamless for my code as I first hoped!

**Comment by PCAssistSoftware** [comment]
Okay - worked it out - it is the same event where the error is occurring when removing that is causing me the error in the first place.

RoboQueue_CollectionChanged

When adding jobs one at a time it fires and triggers the code below to add the job information to my resultsdata, but when adding jobs in bulk via the serializer the event only fires once rather than however many jobs are being loaded which results in wrong information being stored in my resultsdata and thus the error occurring when trying to remove entries

```
 If e.Action = System.Collections.Specialized.NotifyCollectionChangedAction.Add Then

            If Not resultsData.Any(Function(p) p.JobName = e.NewItems.Item(0).Name) Then
                resultsData.Add(New ResultsData() With {.JobName = e.NewItems.Item(0).Name})
            End If

        End If
```

If possible .Deserialize needs to fire the RoboQueue_CollectionChanged event for every job it loads so it then processes them in same way as if you were adding each one manually, then the rest of my code would know no different as to whether jobs are being added manually one at a time, or automatically in bulk using  the serializer.

**Comment by RFBomb** [comment]
>If possible .Deserialize needs to fire the RoboQueue_CollectionChanged event for every job it loads so it then processes them in same way as if you were adding each one manually, then the rest of my code would know no different as to whether jobs are being added manually one at a time, or automatically in bulk using the serializer.

I disagree here. The `NotifyCollectionChangedEventArgs.NewItems` property is an `IList` object. If it were a single item, then I agree that it would need to be raised once per item. But its a list of items that have been added to the collection. Your code on the other hand only  evaluates the first item within the collection ( via index 0 ). 

A better way would be to do a ForEach statement for both Removal and Add events within your handler, as both of those may (or may not) contain multiple items.

**Comment by PCAssistSoftware** [comment]
> A better way would be to do a ForEach statement for both Removal and Add events within your handler, as both of those may (or may not) contain multiple items.

A very good point - which I had completely missed! - thank you.

So you mean something similar to this? or would you do it differently (read better!)

```
 If e.Action = System.Collections.Specialized.NotifyCollectionChangedAction.Add Then

            For Each newItem In e.NewItems
                If Not resultsData.Any(Function(p) p.JobName = newItem.Name) Then
                    resultsData.Add(New ResultsData() With {.JobName = newItem.Name})
                End If

            Next

End If
```

RFBomb:
Yea, that's pretty much what I was recommending to do.

PCAssistSoftware:
Will do final testing tomorrow and then get it merged in - thanks again for all your work on this, lots of useful additions here

PCAssistSoftware:
All appears to be working fine for me, so one last check to ensure you are happy, if so I will click the green button and get it merged :)

RFBomb:
No more changes on this in my end

PCAssistSoftware:
All done - on a separate note I see RoboSharp.Extensions is now on [NuGet ](https://www.nuget.org/packages/RoboSharp.Extensions)and already had 75 download :)

PCAssistSoftware:
Bit confused though as to why NuGet page for RoboSharp shows v1.3.3 as updated 2 days ago, when surely this commit was the one which incremented it to that version and was only just merged......

RFBomb:
> Bit confused though as to why NuGet page for RoboSharp shows v1.3.3 as updated 2 days ago, when surely this commit was the one which incremented it to that version and was only just merged......


 Yea, thats quite concerning. We might want to do another version increment, just to be safe here. I wouldn't think it would push until it was merged in, but maybe one of the settings doesn't quite match the settings from the base nuget thing, so it pushed early.

RFBomb:
@PCAssistSoftware  -- BOTH UPLOADED 2 DAYS AGO.
#184

PCAssistSoftware:
Definitely odd - now merged the other update - thanks for doing that so quickly

tjscience:
Hey, guys. The version showing posted several days ago could have been me figuring out the nuget stuff for the extensions package. I manually deployed the one that failed to test that it was working properly. This was an issue that MS had to fix to be able to publish. Let me know if you see any issues specific to nuget deployment.

RFBomb:
All good, just caught us off guard lol
We incremented again to ensure latest version has all the fixes in it

PCAssistSoftware:
@RFBomb 

Excuse asking this question here, but didn't really want to start a new thread as question was specific to you and though here easiest place to get your attention :)

In my app the "live" status of a copy job is shown in a grid, and in the **RoboQueue_RunCompleted** event I have always then reset the values in this grid to show 0 to copy for files and directories and 0 bytes and set the percentage column to 100% - just so in those rare occasions where my calculations were off a completed job always finished with the grid looking tidy e.g. below

image

I have spent hours today trying to work out why a particular job kept finishing showing the below despite me manually setting the values as I said above.

image

After much diagnosing I have worked out that for some reason the ProgressEstimator_ValuesUpdated event is firing one final time AFTER the RoboQueue_RunCompleted event and it is this which is then changing the values in my grid.

So my question is a) why would it be doing this as I never had issue before and assumed last event to fire would be RoboQueue_RunCompleted as it signified the end of all jobs in the queue? and b) if not then is there easy way in ProgressEstimator_ValuesUpdated for me to check if "job is running or queue is running" or similar to only update my grid if it is and not when it isn't to avoid the problem I now seem to have?

I hope this make sense......   please ask if not

RFBomb:
Nothing should have changed in this regard. 

A) RoboQueue should be cancelling the progress estimator updates, it was coded as such. 

B) no easy way to tell from the event itself. 

Potential solution:
- modify RoboQueueProgressEstimator.CancelTasks to return the task it generates, instead of returning void. This is a bit of a rookie mistake on my part (to be fair I was a rookie at the time of development). 

- Then modify RoboQueue where it calls that method to wait until the Cancellation task completes. This pair of changes ensures that it will always raise the RoboQueue event last.

PCAssistSoftware:
I had just tested an older version prior to all the changes to ensure it was nothing I had broken in my app and to confirm (as you correctly say - and I thought anyway) that nothing has changed regarding this - I think it has always happened, just never noticed, until a specific copy job I was running was giving me those odd results.

I was banging my head against a wall all day on this thinking it was something screwy in my coding, until in the end I simply incremented a counter and output to console in both events so I could see order they were being raised in, and voila realized every single time that ProgressEstimator_ValuesUpdated was firing one final time after RoboQueue_RunCompleted, my console output looks like this from last two runs:

> RoboQueue_RunCompleted:0
> ProgressEstimator_ValuesUpdated:1
> ProgressEstimator_ValuesUpdated:2
> ProgressEstimator_ValuesUpdated:3
> ProgressEstimator_ValuesUpdated:4
> RoboQueue_RunCompleted:5
> ProgressEstimator_ValuesUpdated:6

> RoboQueue_RunCompleted:0
> ProgressEstimator_ValuesUpdated:1
> ProgressEstimator_ValuesUpdated:2
> ProgressEstimator_ValuesUpdated:3
> ProgressEstimator_ValuesUpdated:4
> ProgressEstimator_ValuesUpdated:5
> ProgressEstimator_ValuesUpdated:6
> RoboQueue_RunCompleted:7
> ProgressEstimator_ValuesUpdated:8

Would be great if these code changes could be made at some point as I do a lot of "finalizing" in that event to clear up my form and create logs etc etc, so rely on it being the last event triggered

RFBomb:
Yea, that's a race condition that we just never noticed. 
I have a 6hr flight on Friday I can maybe fix it on.

PCAssistSoftware:
> Yea, that's a race condition that we just never noticed. I have a 6hr flight on Friday I can maybe fix it on.

Thank you, would really appreciate it if you do have time to do it

PR #186, opened 2023-10-25
Update RoboQueueProgressEstimator.CancelTasks

RFBomb:
RE:  https://github.com/tjscience/RoboSharp/pull/183#issuecomment-1777920671

 - Should resolve issue with RoboQueueProgressEstimator.ValueUpdated firing after RoboQueue.RunCompleted

@PCAssistSoftware  I was bored in the hotel room, and I knew this would only take about 20 minutes or so. I think this should resolve your event order issue.

### DIFF
--- a/RoboSharp.Extensions/RoboSharp.Extensions.csproj
+++ b/RoboSharp.Extensions/RoboSharp.Extensions.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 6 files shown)


PCAssistSoftware:
Thank you - really appreciate you doing this so quickly - will test later today when back in front of computer and respond then :)

PCAssistSoftware:
Works like a dream - absolute star, thank you.   If you have no further changes to make on this pull I will get it merged.

RFBomb:
None that I'm aware of

Issue #187, opened 2023-11-23
[Question] Parse Robocopy Parameters


LordPinhead:
Hi,

we use Robocopy for many server applications and we put our parameters for each service into an etcd vault to change the parameters if needed on the fly.
Now I found this little gem and love the idea of getting the feedback of Robocopy in our browser windows, but I don't want to tell my people to forget about the Robocopy parameters and learn new ones.

So how can I use the old rc options / settings and let them get parsed and used? Or is that not done atm, do I have to make a pull and make the parser myself?

RFBomb:
Hello, we must've forgotten to check here with all the holiday hub-bub. 

This was a conundrum I had attempted to created a solution for @PCAssistSoftware, and what I landed on was two solutions:

JobFile - this class when used to save a command will save it as the RCJ file that robocopy likes to store option with. (We actually simply run the command as a save only to generate the file).  `JobFile.Parse` is how you can take that native rcj file and load it back into RoboSharp. 

I believe I have also implemented the feature of robocopy where you can import multiple rcj files. (For this, parse each one, then apply them together to Robocopy) 


---

The other solution is much more recent. This uses `RoboCommandXmlSerializer` to save a collection of commands to a single xml file, whereas the other serializer currently implemented created a new rcj file for each command. 

---

If your existing files are stored in RCJ format (as in saved by robocopy), JobFile.Parse should work. If it's in your own custom format, you would need to create a parser. But `RoboCommandFactory` was designed for this use case, as I use it to generate commands on the fly in my application.

LordPinhead:
I added a class to parse the CMD parameters, so users can migrate to the library without changing their settings or call parameters.
Maybe it´s just for me, but I´ll make a pull request when I have time and add it. 

Thx for the Info nonetheless, it would be great when you would release the change into this main repo.

RFBomb:
@LordPinhead  - Are you suggesting a method to create a RoboCommand by parsing the command prompt parameters? ( ex : ` robocopy "Source" "Destination" /XC /E ` )

LordPinhead:
The last part, "/XC /E /A" etc.
It's a pretty easy parser, I have a lot on my table right now, but I started with the parameters our people use in the company jobs.

It's just a foreach and a switch-case with all the options - I need some time to code down the rest and I will make a pull request later.

RFBomb:
@LordPinhead  -- I have come up with a draft pull request for you to review at your leisure.
 #188 

[...]

PR #188, opened 2024-01-23
Create RoboCommandParser.cs

RFBomb:
This is in reference to #187. This creates a factory object to produce IRoboCommand objects based on an input string such as `robocopy "source" "destination"  /mirror`.

ToDo:
 - Unit tests
 - Higher level logic such as included/excluded files
 - testing extraction of parameters like the Min/Max file age.

### DIFF
--- a/RoboSharp.BackupApp/MainWindow.xaml
+++ b/RoboSharp.BackupApp/MainWindow.xaml
@@ -8,33 +8,56 @@
             <TabItem Header="Options" x:Name="OptionsTab">
                 <Grid>
                     <ScrollViewer Margin="0,0,0,0" VerticalScrollBarVisibility="Auto" Width="Auto" Height="Auto">
-                        <Grid x:Name="OptionsGrid">
-                            <Grid.RowDefinitions>
-                                <RowDefinition Height="10"/>
-                                <RowDefinition Height="Auto"/>
-                                <RowDefinition Height="Auto"/>
-                                <RowDefinition Height="Auto"/>
-                                <RowDefinition Height="Auto"/>
-                                <RowDefinition Height="Auto"/>
-                                <RowDefinition Height="*"/>
-                            </Grid.RowDefinitions>
-                            <Grid.ColumnDefinitions>
-                                <ColumnDefinition Width="173*"/>
-                                <ColumnDefinition Width="540*"/>
-                            </Grid.ColumnDefinitions>
+                        <StackPanel x:Name="OptionsGrid" Height="auto">
+                            <!-- Styles -->
+                            <StackPanel.Resources>
+                                <Style TargetType="StackPanel" x:Key="userInputs">
+                                    <Setter Property="Orientation" Value="Horizontal"/>
+                                    <Setter Property="Margin" Value="10,10,10,0"/>
+                                </Style>
+                                <Style TargetType="TextBlock" x:Key="userInputTxtBlock">
+                                    <Setter Property="Margin" Value="0"/>
+                                    <Setter Property="VerticalAlignment" Value="Center"/>
+                                    <Setter Property="HorizontalAlignment" Value="Right"/>
+                                    <Setter Property="TextWrapping" Value="NoWrap"/>
+                                    <Setter Property="Width" Value="120"/>
+                                </Style>
+                                <Style TargetType="TextBox" x:Key="userInputTextBox">
+                                    <Setter Property="Margin" Value="10,0"/>
+                                    <Setter Property="MinWidth" Value="175"/>
+                                    <Setter Property="MaxWidth" Value="600"/>
+                                    <Setter Property="Height" Value="23"/>
+                                </Style>
+                                <Style TargetType="Button" x:Key="userInputButton">
+                                    <Setter Property="Margin" Value="0"/>
+                                    <Setter Property="Width" Value="75"/>
+                                </Style>
+                            </StackPanel.Resources>
                             <!-- Source -->
-                            <TextBlock Grid.Row="1" Margin="34,0,0,0" HorizontalAlignment="Left" TextWrapping="Wrap" Text="Source" VerticalAlignment="Top"/>
-                            <TextBox x:Name="Source" Grid.Row="1" HorizontalAlignment="Left" Height="23" Margin="34,20,0,5" TextWrapping="Wrap" VerticalAlignment="Top" Width="560" Grid.ColumnSpan="2"/>
-                            <Button x:Name="SourceBrowseButton" Grid.Row="1" Margin="433,20,0,5" HorizontalAlignment="Left" VerticalAlignment="Top" Width="75" Content="Browse" Click="SourceBrowseButton_Click" Grid.Column="1"/>
+                            <StackPanel Style="{StaticResource userInputs}">
+                                <TextBlock Text="Source" Style="{StaticResource userInputTxtBlock}" />
+                                <TextBox x:Name="Source" Width="400" Style="{StaticResource userInputTextBox}"/>
+                                <Button x:Name="SourceBrowseButton" Content="Browse" Click="SourceBrowseButton_Click" Style="{StaticResource userInputButton}"/>
+                            </StackPanel>
                             <!-- Destination -->
-                            <TextBlock HorizontalAlignment="Left" Grid.Row="2" Margin="34,0,0,0" TextWrapping="Wrap" Text="Destination" VerticalAlignment="Top"/>
-                            <TextBox x:Name="Destination" Grid.Row="2" HorizontalAlignment="Left" Height="23" Margin="34,20,0,5" TextWrapping="Wrap" VerticalAlignment="Top" Width="560" Grid.ColumnSpan="2"/>
-                            <Button x:Name="DestinationBrowseButton" Grid.Row="2" Margin="433,20,0,5" HorizontalAlignment="Left" VerticalAlignment="Top" Width="75" Content="Browse"  Click="DestinationBrowseButton_Click" Grid.Column="1"/>
+                            <StackPanel Style="{StaticResource userInputs}">
+                                <TextBlock Text="Destination" Style="{StaticResource userInputTxtBlock}"/>
+                                <TextBox x:Name="Destination" Width="400" Style="{StaticResource userInputTextBox}" />
+                                <Button x:Name="DestinationBrowseButton" Content="Browse"  Click="DestinationBrowseButton_Click" Style="{StaticResource userInputButton}"/>
+                            </StackPanel>
+                            <!-- RoboCommandParser -->
+                            <StackPanel Style="{StaticResource userInputs}">
+                                <TextBlock Text="Command Parsing" Style="{StaticResource userInputTxtBlock}"/>
+                                <TextBox x:Name="ParserText" Width="400" ToolTip="Paste a valid RoboCopy string here to parse it" Style="{StaticResource userInputTextBox}" />
+                                <Button x:Name="ParseTextBtn" Content="Parse" ToolTip="Parse the input string"  Click="ParseTextBtn_Click" Style="{StaticResource userInputButton}"/>
+                            </StackPanel>
                             <!-- Job Name -->
-                            <TextBlock HorizontalAlignment="Left"  Grid.Row="3" Margin="34,0,0,0" TextWrapping="Wrap" Text="Job Name (Optional)" VerticalAlignment="Center" />
-                            <TextBox x:Name="JobName"  Grid.Row="3" HorizontalAlignment="Left" Height="23" Margin="159,5,0,5" TextWrapping="Wrap" VerticalAlignment="Center" Width="175" Grid.ColumnSpan="2"/>
+                            <StackPanel Style="{StaticResource userInputs}">
+                                <TextBlock Text="Job Name (Optional)" Style="{StaticResource userInputTxtBlock}"/>
+                                <TextBox x:Name="JobName"  Width="175" Style="{StaticResource userInputTextBox}"/>
+                            </StackPanel>
                             <!-- Job Options -->
-                            <StackPanel Grid.Row="4" Grid.Column="0" Grid.ColumnSpan="2" Margin="34,10,20,10">
+                            <StackPanel Margin="34,10,20,10">
                                 <Expander Background="LightBlue" Header="Copy Options">
                                     <Grid Margin="10">
                                         <Grid.RowDefinitions>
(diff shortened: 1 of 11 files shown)


RFBomb:
Regex to implement for exclusion parsing:

(?<F>/XF\s+((("[\w\-.*\s]+")|([\w\-*.]+))\s*)+)


(?<xd>/XD\s*
((
(“[\w\\\/$:\-\s]+”) |
((//[a-zA-Z]|[A-Z]:|[^/])([\w\\\/$:]+))
)\s*)+)

RFBomb:
This is now functional and ready for review/merge.  
@PCAssistSoftware  

I believe I've covered all of the options currently provided. Its a lot, so a second set of eyes would be nice. The main file to review is RoboCommandParser.cs, as its the one that has all the instructions for parsing. The other functions are in the ROboCommandParserFunctions.cs, which are all unit tested, so my only real concern is that I forgot to check for an option.

PCAssistSoftware:
Nice one, will try and find sometime this coming week to look over the code and run some tests, hopefully @LordPinhead will also check it out and comment

PCAssistSoftware:
Okay found a few minutes to do some initial testing

Obviously other than just following the PR updates I haven't really looked into the code etc, so my first test was simply based upon using a robocopy.exe command I had recently used so I used the following.  At this point I obviously didn't know if you were expecting to just provide the switches / commands or the whole command, so I included robocopy.exe as well.

`Dim c = RoboCommandParser.Parse("robocopy.exe \\pc-01\c$\SoftwareLib\LegacySoftware C:\test /E /ZB /MIR /MT:128")`

This didn't work

So I tried it again with just robocopy rather than robocopy.exe and this worked perfectly :)

`Dim c = RoboCommandParser.Parse("robocopy \\pc-01\c$\SoftwareLib\LegacySoftware C:\test /E /ZB /MIR /MT:128")`

as did the below where I just provided the commands.

`Dim c = RoboCommandParser.Parse("\\pc-01\c$\SoftwareLib\LegacySoftware C:\test /E /ZB /MIR /MT:128")`

So might be worth covering all basis in case the provided string includes Robocopy.exe perhaps?

Hope that makes sense.

PCAssistSoftware:
More complex command worked fine 

`Dim c = RoboCommandParser.Parse("C:\source\folder\path\ D:\destination\folder\path\ /E /Z /ZB /R:5 /W:5 /TBD /NP /V /XD ""C:\source\folder\path\to\exclude folder""")`

BUT if you have spaces in source and/or destination path then it doesn't work e.g.

` Dim d = RoboCommandParser.Parse("C:\source\folder\path one\ D:\destination\folder\path two\ /E /Z /ZB /R:5 /W:5 /TBD /NP /V /XD ""C:\source\folder\path\to\exclude folder""")`

LordPinhead:
Oh I will check this one out this week and give it a try. Because we use ETCD for such jobs, I have a config entry for the parameters, so I just need them parsed atm. But if I can use the whole line, it's ok too. As long as my co-workers can build jobs on my middleware, I'm happy with it :)

PCAssistSoftware:
Need to be careful with switches such as /NJS /NFL /NDL - which yes it parses okay, but if set to true then going on my previous notes would break apps using RoboSharp e.g. /NFL set to true would break displaying of file progress or file name in your GUI. /NDL breaks logging as doesn't then contain full information and finally and more importantly I believe if /NJS is set to true it breaks statistics completely

RFBomb:
That's a good point about 'robocopy.exe' vs 'robocopy', and is an easy thing to incorporate. I'll even take a look at accommodating paths such as 'D:/robocopy.exe'


I was hesitant about some of the other switches that I'm aware of breaking RoboSharp, BUT those switches are still provided by the library, so I took the position of it's no different than if they set the flag in code, even if it does break things like statistics. I'm simply parsing into an equivalent command.

RFBomb:
Does robocopy actually accept that command with spaced and non-quoted source/destination? Everything I can find says that quotes are required for paths with spaces.

`C:\source\folder\path one\`
`D:\destination\folder\path two\`


Should be 

`"C:\source\folder\path one"`
`"D:\destination\folder\path two"`

LordPinhead:
Pretty much yes, every space marks the end of a command argument in any commandline or shell system

RFBomb:
> /NJS is set to true it breaks statistics completely

This used to be the case, but with my implementation of ProgressEstimator, it can still report something (assuming PE doesn't break due to other switches being set)

PCAssistSoftware:
> I was hesitant about some of the other switches that I'm aware of breaking RoboSharp, BUT those switches are still provided by the library, so I took the position of it's no different than if they set the flag in code, even if it does break things like statistics. I'm simply parsing into an equivalent command.

Fair point

PCAssistSoftware:
> > /NJS is set to true it breaks statistics completely
> 
> This used to be the case, but with my implementation of ProgressEstimator, it can still report something (assuming PE doesn't break due to other switches being set)

Oh okay, I hadn't actually tested it yet, was just looking at the parsing for each switch and comparing with my notes and noted I had put a warning about /NJS from the past

PCAssistSoftware:
> Does robocopy actually accept that command with spaced and non-quoted source/destination? Everything I can find says that quotes are required for paths with spaces.
> 
> `C:\source\folder\path one\` `D:\destination\folder\path two\`
> 
> Should be
> 
> `"C:\source\folder\path one"` `"D:\destination\folder\path two"`

Good point, no it doesn't - but perhaps would be good to have some sort of error handling so if someone did try and parse an invalid string like that it just fails rather than producing wrong commands? just an idea

RFBomb:
> perhaps would be good to have some sort of error handling so if someone did try and parse an invalid string like that it just fails rather than producing wrong commands?

Not opposed to that.
We would have to somewhat strictly define the requirements then of when an exception will throw.  For example, should we require a source/destination? Should we require it begins with 'robocopy' ? What about paths to a specific robocopy? 

`robocopy C:\Some Dir\ D:\SomeDestination`

Would result in a command with source `C:\Some`, destination `Dir\` and file inclusion `D:\SomeDestination`

Obviously right now we can tell that's invalid. But for throwing errors, how do we validate against that? Ensuring that if a source/destination were parsed that both are fully qualified?  If they are not found at all, is that an error?

PCAssistSoftware:
Yes that is a bit more of an awkward one and I suspect not everyone would have the same opinion?

Personally I think it should require that it includes a source / destination
What it begins with should if possible cover a) starting with robocopy, b) starting with robocopy.exe, c) starting without a or b and just starting with source first e.g.

a) robocopy "C:\Some Source\" "D:\SomeDestination"
b) robocopy.exe "C:\Some Source\" "D:\SomeDestination"
c) "C:\Some Source\" "D:\SomeDestination"

LordPinhead:
When you parse the string and the argument you find is not starting with a / , that should work as a check.
i.e. c:\temp\My Documents k:\test
It would be "Source" -> c:\temp\my " destination "Documents" and the first "argument is "k:\test", so it's not starting with the / and should throw an error with the hint of quoting paths.

RFBomb:
I can agree with your ABC examples. Robocopy also requires fully qualified source/destination. So if either path is not fully qualified, throw an error as such.

Robocopy syntax requires that any file filters immediately follow source/destination. There is no /switch for filters, only for exclusions. As such, I'm currently taking anything between the end of the destination and the first / as a file filter. Any thoughts on this aspect?

PCAssistSoftware:
> I can agree with your ABC examples. Robocopy also requires fully qualified source/destination. So if either path is not fully qualified, throw an error as such.

agreed

> Robocopy syntax requires that any file filters immediately follow source/destination. There is no /switch for filters, only for exclusions. As such, I'm currently taking anything between the end of the destination and the first / as a file filter. Any thoughts on this aspect?

yes I agree anything following destination and before first / is a file.filter and this is optional as if you don't specify then RoboCopy just assumes *.*

RFBomb:
The changes have been implemented for review.

PCAssistSoftware:
So far testing going well and works as expected.

Still think we should not parse (or at least warn) for /NFL (LoggingOptions.NoFileList) as when set to true then areas such as showing progress of current file copying don't work correctly

EDIT: or perhaps some way of not including it for those of us who have designed apps which would break with it included - so it covers everyone's needs?

RFBomb:
>Still think we should not parse (or at least warn) for /NFL (LoggingOptions.NoFileList) as when set to true then areas such as showing progress of current file copying don't work correctly 

Well, the options are:
A ) throw an error and stop the whole parsing
B ) Parse and Ignore (no feedback)
C ) Parse as is.

I'm still of the opinion for parse as is, and throw an error only if a major error occurred (such as missing source/destination). 

>or perhaps some way of not including it for those of us who have designed apps which would break with it included - so it covers everyone's needs

There is no rule stipulating that apps the utilize the library must rely on this parser to create commands. Even if they did, it would be up to that developer to ensure that they set up the command correctly for their particular use case. In your example, I would suggest: 

```
dim rc = RoboCommandParser.Parse(input);
rc.LoggingOptions.NoFileList = False;
```

Other consumers may not care about the file list, so to change their entire log for the benefit of something they may not be using doesn't seem correct to me. ( They may not be doing anything with the file list at all, simply using it to start/stop/cancel/track when a command completes )

PCAssistSoftware:
ok

RFBomb:
I have implemented the Parser into the example app so that concerns can be tested using that.

PCAssistSoftware:
> I have implemented the Parser into the example app so that concerns can be tested using that.

Nice one :)

RFBomb:
Updated again, this now has a `ParseOptions()`, `TryParse` and `TryParseOptions` method.
The ParseOptions methods is intended for use with a string of options that does not declare a source/destination. If a source/destination are declared, it will cause an exception. 

This has also been implemented into the example app for testing. I believe this now satisfies all potential uses.

PCAssistSoftware:
> Updated again, this now has a `ParseOptions()`, `TryParse` and `TryParseOptions` method. The ParseOptions methods is intended for use with a string of options that does not declare a source/destination. If a source/destination are declared, it will cause an exception.
> 
> This has also been implemented into the example app for testing. I believe this now satisfies all potential uses.

Yes I noticed the commit and had then been trying it out in the example app - great work (as always!)

PCAssistSoftware:
All working fine from what I have tested, will just wait for @LordPinhead to test / comment and then if all happy can merge

PCAssistSoftware:
@LordPinhead - Hi, could you test this when you get a moment and confirm it works okay for your needs as well? Then I can get it merged in and released.  Thanks

LordPinhead:
I try to test it this week, we are crunchtiming atm with a release of our main ERP and I have severe bugs in our Nextcloud update. I try to get 10 minutes for a test with our example.

RFBomb:
I will be making a tiny change to the new Default RoboCommand factory this morning to return the interface instead of the typed object for a bit more protection. (Prevent consumers messing with the default factory settings, if they want to customize the settings they can instantiate their own instance of the factory)

edit: done

LordPinhead:
I finally had time to change my project to your version. When I tested it, the path had "robocopy" in it, but I did not add the "robocopy" to the start or the string, so it threw me an exception. I added the robocopy to the string, it still threw me an exception. Optionally, the check in TrimRocopy should check if the string starts with robocopy and not in the path, that´s why I split the arguments and paths. When somebody has has robocopy in his destination path, this would be a problem (I remember a coworker that had this). For me, it was just in a local test.

Otherwise, it works like a charm :) Thx for the work,

PCAssistSoftware:
@LordPinhead - thanks for testing

@RFBomb  - yes I can replicate the scenario that LordPinhead mentions and also get an error e.g.

 `Dim x = RoboCommandParser.Parse("robocopy.exe d:\SoftwareLib\RoboCopyTest C:\test /E /ZB /MIR /MT:128")`

image

LordPinhead:
Also, would single quoted paths work in CMD? I have to check that, I think it´s working but it´s not working with the parser.

RFBomb:
Adding in single-quoted paths is possible, but unfortunately I just tested it out, you can create a new folder whose name is a single quote : `'`. 
If the desire is there, we can try to patch it in, but given its a valid character in a file/folder name, whereas double quotes are not, I think sticking to double quotes is the better standard.

I had not considered someone having a path with the name 'robocopy' in it when I wrote that bit of code .

edit: A workaround is still likely needed for that issue. 
` d:\robocopy` will execute the file `d:\robocopy.exe` if it exists. that was the reason I was throwing the error 
Now, if you pass in the following : 
`c:\myfolder\robocopy d:\mydestination` it will still error out, but this time with an unable to parse source/destination error. This is because `c:\myfolder\robocopy` could start the process via command prompt, so it has to be treated as such. 

workaround: either put 'robocopy' at the start of the input text. OR ensure that the source/destination paths end in a `\` character. 

ERRORS -- `c:\myfolder\robocopy d:\mydestination`
        OK -- `c:\myfolder\robocopy\ d:\mydestination`
        OK -- `robocopy c:\myfolder\robocopy d:\mydestination`

RFBomb:
single quotes do not work on cmd - they are treated as a filename. In this case, it assumed I had a folder path of `c:\'c:\rst'\` when my inputs was `'c:\rst'`

image

PCAssistSoftware:
Thanks @RFBomb - I totally agree with all your logic above, and have tested your changes and all works for me.  If @LordPinhead you could test and comment then if we are all happy I will merge this PR in

LordPinhead:
@PCAssistSoftware  I´m happy, you can merge it, I like the implementation. A big thanks @RFBomb :)

PCAssistSoftware:
@LordPinhead - thanks for that

@RFBomb - just checking you don't want to change anything else, if not I will merge it :)

PCAssistSoftware:
Hi @RFBomb 

Found an issue if parsing a command line which has no source / destination then it seems to ignore the file filter supplied e.g.

Dim w = RoboCommandParser.Parse("""" *.docx /E")
      Dim x = RoboCommandParser.Parse(""""" """" *.docx /E")
      Dim y = RoboCommandParser.Parse("robocopy.exe """" *.docx /E")
      Dim z = RoboCommandParser.Parse("robocopy """" """" *.docx /E")

Whether I use "" or "" "" for blank source / destination it still ignores the *.docx for file filter

Any thoughts please

RFBomb:
ParseOptions should work if you know that there is no source/destination in the string. 

Otherwise we can double check it with some unit tests. Some tests should already be set up for that, but I'll review it

PCAssistSoftware:
Hadn't tried ParseOptions, as in my app I wanted user to be able to enter a whole command either with or without source / destination, so app wouldn't know until parsing it that no source / destination had been supplied e.g. using "" instead.

Thank you

PR #191, opened 2024-02-16
Fix robo command ambiguity

RFBomb:
A long time ago, when I was a wee beginner, I added a flaw into the constructors of RoboCommand : Ambiguity.
This was primarily caused by the introduction of the 'stopIfDisposing' optional bool.

This PR adds several new constructor overloads for RoboCommand to allow the compilers to more easily find their 'best match' given the parameters. Overall functionality is unchanged, these overloads just ensure that a command that specifies source/destination has an exact matching constructor, instead of being ambiguous until you add a third parameter.

### DIFF
--- a/RoboSharp/RoboCommand.cs
+++ b/RoboSharp/RoboCommand.cs
@@ -34,6 +34,13 @@ namespace RoboSharp
             Init();
         }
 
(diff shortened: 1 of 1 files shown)


PCAssistSoftware:
That all makes sense.  If you can answer my final query in 189 in case that needs a change, then I will merge both of these in.

PR #193, opened 2024-02-26
RoboSharp.Extensions - Async File Copier

RFBomb:
So for the past few weeks I've been noodling with this, and after a brief review of the Extensions package as it stood, I decided to place my work here. The extensions package already contains references the the Kerner32.dll file for handling SymLinks. Since it is windows specific already, I'm adding in the support for CopyFileEx, which resides within that same system dll. 

This PR provides various functionality for wrapping CopyFileEx and its MoveFile equivalent. This allows native Progress Reporting methods for file copy/move operations for use with the custom IRoboCommand objects. CopyFileEx can only run Synchronously, so all Async methods that utilize it wrap it inside of a `Task.Run()`. 

In order to decouple the caller from the copy operation various methods have been introduced into the `FileFunctions` class, such as using a Progress<double> object to report progress. A new `FileCopier` class has been introduced that extends the existing `FilePair` class to simplify calling of these functions for custom commands.

As an aside, part of the reason for combining this here instead of another nuget package is that I believe under the hood robocopy utilizes CopyFileEx for its progress reporting and restartable mode, as these are also available within that call's api.

The FileCopier class intentionally does not do any retry logic, as that should be implemented within the custom command if desired. It does provide a ProgressUpdated event that creates the appropriate event args though. The `RoboSharp.CopyOptions` object directly translates some of its booleans to the enum that CopyFileEx uses, so an extension method has been added into the Helpers namespace to allow for this translation.


PCAssistSoftware:
Sounds good, can you add a couple of code samples here to demonstrate how to use it?  then I can have a play with it when I get a moment

RFBomb:
Yea I'm working on that side of things. 
One of the things I'm sorta debating though is if this is even a 'good idea'. As CopyFileEx (while simply to use and likely wont be deprecated any time soon), is superceded by CopyFile2 apparently. So I'm wondering if the PR should instead have a base implementation laid out instead, and let the caller decide to plug in their own progress-reporting copier. 

CopyFileEx has been around since XP. CopyFile2 requires Windows8 or newer (most development is this scenario, but potentially not all, legacy systems exist)

PCAssistSoftware:
Yes awkward one that but can't see CopyFileEx going anywhere for a long time as used by so many products and other functions.  To be honest I had heard of that but never of CopyFile2 so that says a lot.

RFBomb:
Yes I discovered CopyFile2 while perusing the web for CopyFileEx examples, and stumbled upon a few threads that recomended 'if an *Ex function is superseded you should use the new method instead.'

The major difference between them that I understand is that CopyFile2 uses an IntPtr struct to pass in all arguments, and has already gone through several revisions of that struct, adding some additional capabilities to it (the CopyFile2 call depends on the struct you pass in). Whereas CopyFileEx has a specific signature that doesn't allow for that flexibility.

I think for the needs here, CopyFileEx suits the needs just fine. But I'm going to write it in such a way that CopyFile2 (or just filestreams) can be used if the consumer wishes to implement it.

RFBomb:
@PCAssistSoftware 
This is coming along fairly well. Unfortuantely I did run into a bit of inconsistency:
- On my home computer, the Unit Tests run fine!
- On my work computer, they fail.

image

I was hoping you could download this PR and run the `CopyFileEx_CopyFile` test yourself to see if it also fails. 
This seems to be an inconsistency in CopyFileEx invoking across systems.

PCAssistSoftware:
Worked fine for me.

image

Let me know anything else I can test for you.

PCAssistSoftware:
for info the one which fails for me is CopyFileAsyncTest

image

RFBomb:
I HAVE A CONUNDRUM.

image

VS2022 PASSES the unit test and VS2019 FAILS it!!!
what in the world........

PCAssistSoftware:
That is so odd?!?

RFBomb:
So obviously, if I can reliably reproduce these results on two different computers. it is a crapshoot as to what the behavior of the nuget package that appveyor produces will result in. (But that's why I write the test). 

I have created a discussion post on the.net/runtime Because I don't know where else to report the issue. 

Other than this hiccup, This PR is coming along fairly well I think. It still has a long way to go because I intend to incorporate some of the changes into the test app for demo/test purposes. Much of that is already marked up but I have to finalize everything.

PCAssistSoftware:
Sounds good.  Will be interested to hear if you get any response to help explain it.

RFBomb:
Alright, I did get a response, implemented the change, and its now unit testing fine on both VS2022 and 2019 on my pc. It is now back to what I think I had originally, so not sure what I did wrong in the interim, but its working again so thats that. 

More unit tests have been added and things have been refined. This is now 'functional'.
The required changes to the UI have not yet been implemented, so its not ready for merging yet.

As an overview of usage:

`FileCopyExFactory` and `StreamedCopierFactory` are the two new factory objects to produce `IFileCopier` objects. 
These `IFileCopierFactory` objects are usable for custom implementations, such as the new `RoboBatchCommand`. 

`RoboBatchCommand` was my first run at a custom IRoboCommand for my application, and it works fairly well. Basically provide it a collection of source/destination pairs and it will run through the collection, copying each to their respective destination. This allows to say grab 20 files from 20 seperate folders into either one destination or up to 20 ( 1 destination per source ), and run as a single IRoboCommand

```
var batch = new RoboBatchCommand(new StreamedCopierFactory()); // this is platform agnostic - hello linux :)
batch.AddCopier("SomeSource", "SomeDestination");
batch.AddCopier("SomeOtherSource", "SomeOtherDestination");
await batch.Start()
```

PCAssistSoftware:
Excellent!

Just run all tests here and I am getting 3 failures as per screenshot below - all have same error relating to attempting to read or write protected memory.

image


This is looking really great, fantastic work :)

RFBomb:
Strange. I bet it's because the path is inside of your OneDrive folder. Mine is C:\Repos. 
Since it's iterating across the test in <500ms one drive probably is picking up the new file, holding it, then the issue occurs.

PCAssistSoftware:
> Strange. I bet it's because the path is inside of your OneDrive folder. Mine is C:\Repos. Since it's iterating across the test in <500ms one drive probably is picking up the new file, holding it, then the issue occurs.

moved it outside of my OneDrive and now get 6 failures....??!?!?

image

PCAssistSoftware:
Moved it to the root of another drive so path was shorter to rule out path length and got 4 failures.  Very odd..

image

RFBomb:
Not that it helps here, but I ran successfully ><
image

The last commit changed where I'm writing the files for the test, maybe that helps? I really don't know why it would be having that error

PCAssistSoftware:
Just tried last commit and getting 5 failures, 1 for the same error as before "System.AccessViolationException: Attempted to read or write protected memory" and the other 4 for a new error of "System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\Users\darre\AppData\Local\RoboSharpUnitTesting\auw35chf.sbu'."  (with obviously a different filename for each of the 4 failures)

image

RFBomb:
Passing in VS2019 and failing in VS2022. debugging now...

PCAssistSoftware:
Success!!!

image

RFBomb:
I find it very odd that one of the tests passed and other didn't in VS2022, more specifically I find it strange that VS2022 threw the error and VS2019 was not. But it is now unit testing successfully in both environments for me.

OF COURSE APPVEYOR FAILED. ><

PCAssistSoftware:
> I find it very odd that one of the tests passed and other didn't in VS2022, more specifically I find it strange that VS2022 threw the error and VS2019 was not. But it is now unit testing successfully in both environments for me.
> 
> OF COURSE APPVEYOR FAILED. ><

yes as per previous reply all fine here in 2022 as well.   All very odd though

RFBomb:
Alright so to summary all the issues that were causing discrepancy from 2019 to 2022 :

- P/Invoking is picky! (as it should be)
- The signature for the call to CopyFileEx was wrong initially in that I was using `ref int pbCancel` instead of `ref bool pbCancel`. But I had changed it to `bool pbCancel`. This resulted in the protected memory exceptions vs2022 was reporting (can't change a constant, so error). Changing it back to `ref bool pbCancel` resolve this exception.
- parameter `lpData` was originally an `IntPtr` but troubleshooting the above issue I had changed it to object. vs2022 did not care about this, as the only usage was an IntPtr and as such it compiled it like one. But VS2019 compiled into an `object`, which the P/Invoke did not like, and the result was cutting off the flag parameter. Changing this to the correct `IntPtr lpData` resolved the issue. 
- Eventually resolving the signature to be correct resolved all tests.

RFBomb:
#Alright, @PCAssistSoftware I overhauled this (again), but this time with a force-push, bringing it from 60+ confusing noodles of commits to 16 (relatively) organized ones. I'm quite satisfied from my own cursory review here, but will need to re-look at my consuming library to double check. That wont happen until next week at the earliest, probably later. 

But feel free to poke around, it should be a bit more organized than previously. Note that many things were re-arranged, and Windows-Specific functionality is now contained in the `RoboSharp.Extensions.Windows` category (CopyFileEx and FileFunctions.MoveWithProgress)

As a note, this is MOSTLY the same as capability-wise as previous, but much smaller number of commits instead of piece-meal many times over.

RoboSharp.Extensions/RoboBatchCommand.cs
```diff
@@ -243,6 +242,10 @@ public override Task Start(string domain = "", string username = "", string pass
                         resultsBuilder.ProgressEstimator.SetCopyOpStarted();
                         queue.Add(copyTask);
                     }
+                    else
```
RFBomb, line 245: Remove these lines

RoboSharp.Extensions/RoboBatchCommand.cs
```diff
@@ -230,7 +230,6 @@ public override Task Start(string domain = "", string username = "", string pass
                     //if (copier.Parent.ProcessedFileInfo is null)
                     //    copier.Parent.ProcessedFileInfo = new ProcessedFileInfo(Path.GetDirectoryName(copier.Source.FullName), FileClassType.NewDir, this.Configuration.GetDirectoryClass(ProcessedDirectoryFlag.NewDir), 1);
 
-                    resultsBuilder.AddFile(copier.ProcessedFileInfo);
```
RFBomb: Add this line back in, delete lines 245-248
Come up with fix to prevent logging the file a second time (a dictionary? A boolean field part of the interface? not sure) in the results builder

PCAssistSoftware:
HI @RFBomb  - just acknowledging receipt of your above message.  Snowed under with work at present but will try and take a good look at it next week :)

RFBomb:
Appreciated. This work was part of an over-arching refactoring of my other project, so I still have yet to complete that, which is when I will be doing my own more in depth testing

PCAssistSoftware:
Can you just give me a quick overview of what new commands are added by your PR and how I would use them, just want to ensure I am testing all the right areas for you.  Thanks

RFBomb:
The high-level summary is that this introduces IFileCopier, and two implementations of it (streamed copier and copyfileX). 

The idea  is that you pass one of those factories into the new batch command and from there you add in your custom batch of files ( file one destination one, file two destination, two, etc.) in the form of IFilePair objects, or using the add methods. 

When you start the command it will take the files specified and copy them. The idea with this is not so much for large server to server transfers where RoboCopy is likely more efficient, but for local transfers where you specify a couple files, Maybe from five or six different directories and copy them all to some destination directory, or even into wildly different locations. 

The app that I have uses this to copy various files off of the c drive and onto a specialized folder structure on a USB. so not all of the files exist within the same source directory, nor do they all go into the same destination directory, it is just a batch of files that need to be collected, copied to some destination, and they will copy one after another

RFBomb:
Also, the current name is RoboBatchCommand, and the more I thought about it the more I dislike it. It feels too similar to RoboQueue. I'm thinking of shortening it as simply `FileBatch` or maybe `BatchCommand`. 

Also, that object currently does not handle copying directories as a whole, Just files. So I was contemplating Support for directories. But then again, that is where you could use a RoboQueue that contains a file batch and a RoboCommand to perform that operation (which I personally think is probably the cleaner method, leave the file batch file batch, and a directory to a proper RoboCommand)

PCAssistSoftware:
Thanks, had an initial look over it and all seems to work fine as far as I can see from my basic testing and looking through the code.  Is a really useful addition!

I agree re changing the name, BatchCommand stood out for me.

PCAssistSoftware:
P.S. If there are any specific areas you need testing at any time, just let me know

RFBomb:
Holy cow, that one test that kept failing was a hurdle. (I knew it was a race condition due to a small file size, but atleast it finally passed). 

I know I said it before, but I'll say it again: this is pretty much done.  Since the last comment, various optimizations to the core library (specifically handling string generation) have been made, most of which were benchmarked locally. 

CopyFileEx and IFileCopiers are at a pretty good spot I think. I intend to take a look at the GUI testing app this afternoon for easier testing by hand.

PCAssistSoftware:
Sorry for delay in responding, been one of those weeks!.  All looking good, let me know if you want anything testing (if so then please give me some pointers on usage) :)

RFBomb:
No worries at all, I've been busy on other jobs and vacation for a bit, which is why no progress for a few weeks. 

As far as usage goes, I'm reworking the BackupApp to showcase it in a new tab. The ViewModel will show how to use it within an app.

RFBomb:
Alright, this last commit is a huge OVERHAUL of the entirety of the BackupApp. 
- I was having some trouble reusing some components, so I converted the entire thing into the MVVM style of UI. 
- The various controls now have each have a 'ViewModel' within the viewmodels folder that controls the interaction with the UI. This means the code is more segmented for its use, instead of all bundled into the main window's code-behind.
- Also added was a new tab for the BatchCopy to showcase that functionality and expose it for testing.


I did some brief testing, and all seems to be working from what I could tell. When you get a chance, please poke around the new backup app, I will of course tweak anything as needed.

PCAssistSoftware:
Great work (as always), just had a very quick look at the new BackupApp, one quick fix, when it launches can it open to the correct size to fit all content on first tab, currently (for me at least) it opens looking like below and you have to drag it larger to see all content.

I will spend some time tomorrow checking out the new BatchCopy functionality and get back to you :)

image

RFBomb:
Yes I noticed that as well, and intend to fix it, should be pretty simple. There are a few other minor alignment tweaks I want to do as well. I was just focused on functionality for the overhaul. 

It pretty much all needs to be tested, as it was overhauled entirely, so take your time, I'm in no rush

PCAssistSoftware:
No problem at all, will thoroughly test it all and report back

PCAssistSoftware:
Will append any issues found to this post.

1) If you click Load Job, select a previously created .RCJ file and then click Start Job - it hesitates for a moment and then just closes the window.

2) If you click Load Job and then click add to Queue, again it hesitates for a moment and just closes the window.

For both 1 and 2 the crash occurs at same point in code - error details are below:

image

RFBomb:
Above items were resolved.

PCAssistSoftware:
GUI size now perfect.

1) Can now load job from file and click start job - but job doesn't actually run?

2) adding job to queue and running via "start job queue" works fine, 

3) selecting job in queue, running that job queue and then clicking remove selected causes crash - can remove fine BEFORE you run the job queue, but not afterwards

RFBomb:
1 -- I was able to load and start the job without issue?
2/3 -- resolved.

PCAssistSoftware:
Thanks, ignore 1 as now working fine, so not sure what was happening with that.  Will continue testing

PCAssistSoftware:
Extensions.BatchCommand tab:

1) First thing I did was to type a source file path in box rather than clicking select file - so first failure was that add to batch doesn't enable unless you use select file button.  Can this be changed so it enables when manually entering path?  not to worry if not, was just a habit for me to type path as I knew it rather than browse for it.

2) Tested using StreamedCopier and CopyFileEx to copy an ISO from one folder to another with new file name - both worked as expected, for me CopyFileEx was much quicker, but I am doing this locally on same drive, so assume would notice more differences over a network link perhaps? or when copying multiple files?

RFBomb:
@PCAssistSoftware - I would expect CopyFileEx to be a bit faster based, as its relying on native system calls, whereas StreamedCopier uses a FileStream to perform the asynchronous copy. I introduced the StreamedCopier as a base alternative that could work cross-platform should someone need it. 
No idea when it would be faster. But this app now can run in .net8 as well, which might be marginally faster since it uses a different overload.

PCAssistSoftware:
That make sense

RFBomb:
This last commit fixes net8 crashing on startup due to the VersionManager lookup.
But more importantly it puts the BufferSize adjustment on the backup-app's UI. 

I tested using a 548MB file, Copying from the source file path, to a new file in same directory. The drive was an NVME4 drive. 
CopyFileEx : 0.217s
StreamedCopier (default buffer size of 81920) : ~30s with  - there were also some slowdowns and brief pauses as this copied.
StreamedCopier (Buffer size of 2MB) : 0.317s with  - no slowdowns or pauses, nearly comparable to CopyFileEx

The 81920 came from here : https://learn.microsoft.com/en-us/dotnet/api/system.io.filestream.copytoasync?view=net-8.0

PCAssistSoftware:
I have tested every single button and option multiple times for all methods both in the net48 and net8.0 versions of the BackupApp and all appear to be working fine :)

PCAssistSoftware:
Let me know when you would like me to merge this in :)

RFBomb:
All right, I just spent the last two days updating the symbolic link support and validating it.

I May write a unit test for them tomorrow, but they do require admin privileges for creation, which makes the unit aspect tricky. But resolving the target should work without admin.

(Well, one of two works without admin. The other reports back the link's path instead of following it through-- but will follow to target when called as admin. I left this as an explicit call, instead of extension method). 

As for merging, I'm now able to finalize the rework of my library that will be the primary consumer of these changes. I want to hold off merging this until that 'real world use case' has been tested, as it will let me work any kinks out as a real consumer of these changes.

PCAssistSoftware:
> All right, I just spent the last two days updating the symbolic link support and validating it.

:)

> I May write a unit test for them tomorrow, but they do require admin privileges for creation, which makes the unit aspect tricky. But resolving the target should work without admin.
>
> (Well, one of two works without admin. The other reports back the link's path instead of following it through-- but will follow to target when called as admin. I left this as an explicit call, instead of extension method).
> 
Can you not get it to prompt to elevate to run the test or something similar? e.g. like you would do with an app by modifying / adding an app.manifest (application manifest file" and setting requestedExecutionLevel to requireAdministrator? That is what I do to all my apps using RoboSharp anyway as it needs admin rights to run properly anyway

> As for merging, I'm now able to finalize the rework of my library that will be the primary consumer of these changes. I want to hold off merging this until that 'real world use case' has been tested, as it will let me work any kinks out as a real consumer of these changes.

That makes perfect sense.  Happy to test anything else needed in the meantime.  Do any of the changes in this PR affect or change anything in relation to how RoboSharp currently works or are all of the changes mainly related to the batch / async file copier side?  just checking as quite happy to test this version in all my apps if anything has changed and needs testing.

RFBomb:
The core functionality of RoboSharp should remain unchanged. There were a couple fixes and optimizations put in though. But the vast majority of this PR is reworking and reorganizing the Extensions, to introduce the `IFileCopier` and `BatchCommand`

So while I definitely wouldn't say no to proving it out before releasing to production, I don't anticipate any difference, even if you just drop the dll in place.

RFBomb:
Just finished writing the unit tests for Symbolic Links, all are working on my computer!
These MUST be run as admin (start Visual Studio using admin), otherwise they will fail. Since this is a unit test project and not an executable, there isnt and app manifest to request access in. They will simply fail if user does not have the permissions to create the links.

By calling the native methods, this actually expands past the .Net6 provided code, as the native methods allow for creation and resolution of relative path links, whereas the FileSystemInfo.CreateAsSymbolicLink does not (well, it wont do it automatically, maybe it will work if you calculate the path yourself and put it in). 

It also looks like appveyor was able to pass the tests as well, which is great

PCAssistSoftware:
All tests passing for me as well :)

PCAssistSoftware:
> The core functionality of RoboSharp should remain unchanged. There were a couple fixes and optimizations put in though. But the vast majority of this PR is reworking and reorganizing the Extensions, to introduce the `IFileCopier` and `BatchCommand`
> 
> So while I definitely wouldn't say no to proving it out before releasing to production, I don't anticipate any difference, even if you just drop the dll in place.

Just been doing some tests with my existing apps just to double check.

Initially just dropped in the new DLL but got an error first time I tried to run, but after rebuilding my project using the new DLL it works fine, so not sure what happened or why but just wanted to mention it anyway just in case.

"MissingMethodException: 'Method not found: 'RoboSharp.Interfaces.IStatistic RoboSharp.Interfaces.IRoboCopyResultsList.get_FilesStatistic()'.'"

RFBomb:
Ah, that one did change. It works after the recompile because it was 'technically' removed, but regained it due to now inheriting from the IResult interface. So dropping the new version in wont work if those three IStatistic properties are utilized on IResultList.

PCAssistSoftware:
Thanks for clarifying for me.  Glad I tested it now.  Everything else seems to be fine, will do final testing later in my other apps.

PCAssistSoftware:
Hi, just checking in to see how your 'real world use case'  is going with this PR.  I am not currently using any of the new File Copier methods but I have been using this versions DLL in all my apps and had no problems at all.

RFBomb:
That's great to hear. I've been pushed around with some other projects and had to put this down. I should be able to re-review my WIP library that consumes this within next week or two. I think it had a a few minor tweaks that weren't committed to this, but I have to review. (All here looks fine until you put it into real world, so I'm just making sure I'm done before letting it rip)

PCAssistSoftware:
Excellent, no problem at all, whenever you have time, I know exactly what you mean!

RFBomb:
OK, super busy last few weeks, but I found a chance to review. I am quite confident this is complete as of the last push I just performed. Minor tweaks to the symbolic link functionality (and I did some testing outside the repo). so this should be it.

PCAssistSoftware:
Okay if you are sure then change it from draft and I will get it merged in :)

Ignore that - now changed from draft status (didn't realize all I had to do was click review!), so will merge it

PCAssistSoftware:
for some reason we have a build failing message on main page and nuget packages for this release haven't been created - any ideas?

RFBomb:
It built fine, @tjscience has to update the nuget API key. 

I recently got an email saying the same thing, so they probably just did some Maintenance and are requesting everyone refresh their keys.

Error at end of build:
> Error publishing package. NuGet server returned 403: The specified API key is invalid, has expired, or does not have permission to access the specified package.

PCAssistSoftware:
Thanks, only did quick look and hadn't noticed the error.

PCAssistSoftware:
@RFBomb - API key has been sorted by Terry and v1.6.0 is now on NuGet

Issue #197, opened 2024-03-08
Add "include modified" selection option


kiwiwings:
Please provide the "include modified" (/im) option in the selection options.

I've adapted the included tweaked options, but I haven't got a test case yet, so it's not something to provide as patch.

PCAssistSoftware:
I will take a look at adding this

PCAssistSoftware:
**Notes**

The functionality of the /IM switch can be tested as follows:-

1) Have a source and destination folder containing the same file
2) Use [BulkFileChanger](https://www.nirsoft.net/utils/bulk_file_changer.html) (or similar) to change the created time of the source file
3) Running Robocopy without /IM switch will find nothing to copy, but running it with /IM switch will find the one file with different created time to copy

PCAssistSoftware:
Implemented as requested in #198

PCAssistSoftware:
> **Notes**
> 
> The functionality of the /IM switch can be tested as follows:-
> 
> 1. Have a source and destination folder containing the same file
> 2. Use [BulkFileChanger](https://www.nirsoft.net/utils/bulk_file_changer.html) (or similar) to change the created time of the source file
> 3. Running Robocopy without /IM switch will find nothing to copy, but running it with /IM switch will find the one file with different created time to copy

I will add to this that it also seems to be related to the archive bit being set or not.  The Microsoft documentation and the documentation on page below seem to vary on the description for the switch, and in testing I found that just changing the time as above wasn't always working, but if I cleared the archive bit it was.

https://ss64.com/nt/robocopy.html

PR #198, opened 2024-03-10
Start implementing include modified /IM switch

PCAssistSoftware:
Start implementing include modified /IM switch as per #197

### DIFF
--- a/RoboSharp/SelectionFlags.cs
+++ b/RoboSharp/SelectionFlags.cs
@@ -36,11 +36,11 @@ namespace RoboSharp
         IncludeSame = 256,
         /// <inheritdoc cref="SelectionOptions.IncludeTweaked"/>
         IncludeTweaked = 512,
-        /// <inheritdoc cref="SelectionOptions.IncludeModified"/>
-        IncludeModified = 1024,
         /// <inheritdoc cref="SelectionOptions.OnlyCopyArchiveFiles"/>
-        OnlyCopyArchiveFiles = 2048,
+        OnlyCopyArchiveFiles = 1024,
         /// <inheritdoc cref="SelectionOptions.OnlyCopyArchiveFilesAndResetArchiveFlag"/>
-        OnlyCopyArchiveFilesAndResetArchiveFlag = 4096,
+        OnlyCopyArchiveFilesAndResetArchiveFlag = 2048,
+        /// <inheritdoc cref="SelectionOptions.IncludeModified"/>
+        IncludeModified = 4096
     }
 }
--- a/RoboSharp/SelectionFlags.cs
+++ b/RoboSharp/SelectionFlags.cs
@@ -36,9 +36,11 @@ namespace RoboSharp
         IncludeSame = 256,
         /// <inheritdoc cref="SelectionOptions.IncludeTweaked"/>
         IncludeTweaked = 512,
+        /// <inheritdoc cref="SelectionOptions.IncludeModified"/>
+        IncludeModified = 1024,
         /// <inheritdoc cref="SelectionOptions.OnlyCopyArchiveFiles"/>
-        OnlyCopyArchiveFiles = 1024,
+        OnlyCopyArchiveFiles = 2048,
         /// <inheritdoc cref="SelectionOptions.OnlyCopyArchiveFilesAndResetArchiveFlag"/>
-        OnlyCopyArchiveFilesAndResetArchiveFlag = 2048,
+        OnlyCopyArchiveFilesAndResetArchiveFlag = 4096,
     }
 }
--- a/RoboSharp/JobFileBuilder.cs
+++ b/RoboSharp/JobFileBuilder.cs
@@ -431,6 +431,7 @@ namespace RoboSharp
                 ExcludeOlder = Flags.Any(flag => flag.Success && flag.Value == SelectionOptions.EXCLUDE_OLDER.Trim()),
                 IncludeSame = Flags.Any(flag => flag.Success && flag.Value == SelectionOptions.INCLUDE_SAME.Trim()),
                 IncludeTweaked = Flags.Any(flag => flag.Success && flag.Value == SelectionOptions.INCLUDE_TWEAKED.Trim()),
+                IncludeModified = Flags.Any(flag => flag.Success && flag.Value == SelectionOptions.INCLUDE_MODIFIED.Trim()),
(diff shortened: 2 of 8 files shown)


PCAssistSoftware:
@RFBomb - have started making changes to implement include modified /IM switch as per request in #197

Would appreciate your input and checking of this to make sure I have a) done it correctly and b) also not sure what changes (if any) may be needed in relation to below:

ProgressEstimator.cs   - not sure if we need to handle tweaked in any special way like you do IncludeTweaked around line 453 approx.

RoboSharpConfiguration.cs  line 278 approx. (and related code in ProcessedFileFlags.cs) – do we need to do similar handling re LogParsing for include modified?

RoboSharp/SelectionFlags.cs
```diff
@@ -36,9 +36,11 @@ public enum SelectionFlags
         IncludeSame = 256,
         /// <inheritdoc cref="SelectionOptions.IncludeTweaked"/>
         IncludeTweaked = 512,
+        /// <inheritdoc cref="SelectionOptions.IncludeModified"/>
+        IncludeModified = 1024,
```
RFBomb: This potentially breaks previous jobs!
- If any consumers were using `OnlyCopyArchiveFiles` or `OnlyCopyArchiveFilesAndResetArchiveFlag`, and happened to be also using serialization, that would get serialized to the 1024 value. Upon updating the library, it would read in IncludeModified instead of `OnlyCopyArchiveFiles`.
- Any new values should be appended to the end of the enum. (Location of the value in the list doesn't matter, but the assigned value does)

Recomendation: 
IncludeModified = 4096
OnlyCopyArchiveFiles = 1024
OnlyCopyArchiveFilesAndResetArchiveFlag = 2048

RFBomb (changes requested):

RFBomb:
From the docs (posting for reference) 

> <html><body>

/im | Include modified files (differing change times).
-- | --
/is | Includes the same files. Same files are identical in name, size, times, and all attributes.
/it | Includes "tweaked" files. Tweaked files have the same name, size, and times, but different attributes.


</body>
</html>

> Modified files classification applies only when both source and destination filesystems support change timestamps, such as NTFS, and the source and destination files have different change times but are otherwise the same. These files aren't copied by default. Specify /IM to include them.

---
Most of your changes look good, except on the enumeration of SelectionFlags

Ill take a peak at PE and Configuration momentarily

PCAssistSoftware:
Have resolved the SelectionFlags enumeration - wrongly assumed they should be in order - thank you for advising

RFBomb:
I believe that both Configuration and ProgressEstimator will have to be modified accordingly: 

For configuration, I think we will likely need something for it:
( Note that i'm assuming the default for en-us is `modified`, but only running the command and seeing the inclusion in the log can verify this )

```
/// <summary>
/// Log Lines starting with this string indicate : File was included by <see cref="SelectionOptions.IncludeModified"/> filters
/// </summary>
public string LogParsing_ModifiedInclusion
{
    get { return modifiedIncludedToken ?? GetDefaultConfiguration().modifiedIncludedToken ?? "modified"; }
    set { modifiedIncludedToken= value; }
}
private string modifiedIncludedToken;
```

ProgressEstimator : 
```
case true when currentFile.FileClass.Equals(Config.LogParsing_ModifiedInclusion, StringComparison.CurrentCultureIgnoreCase):  //IncludeModified
                            SkippedOrCopied(currentFile, !Command.SelectionOptions.IncludeModified);
                            break;
```

I also believe that once this is put in, Extensions should add support for it. But I can figure that out as part of my draft PR.

PCAssistSoftware:
> Note that i'm assuming the default for en-us is `modified`, but only running the command and seeing the inclusion in the log can verify this 

That is a correct assumption looking at my test logs

PCAssistSoftware:
@RFBomb - thanks for that, have made suggested changes, please double check I have located them in correct place etc.

RFBomb:
All looks good to me.

PCAssistSoftware:
> > Note that i'm assuming the default for en-us is `modified`, but only running the command and seeing the inclusion in the log can verify this
> 
> That is a correct assumption looking at my test logs

@RFBomb 

Correction, I was incorrect, it actually appears in log with a capital M as per below - is it case sensitive in the code? do I need to change line 292 of RoboSharpConfiguration from "modified" to "Modified" ?

```

-------------------------------------------------------------------------------
   ROBOCOPY     ::     Robust File Copy for Windows
-------------------------------------------------------------------------------

  Started : 10 March 2024 19:25:24
   Source : C:\Users\darre\OneDrive\Desktop\source\
     Dest : C:\Users\darre\OneDrive\Desktop\destination\

    Files : *.*

  Options : *.* /L /DCOPY:DA /COPY:DAT /IM /R:1000000 /W:30

                           1    C:\Users\darre\OneDrive\Desktop\source\
            Modified               1.0 m        v750.pdf

               Total    Copied   Skipped  Mismatch    FAILED    Extras
    Dirs :         1         0         1         0         0         0
   Files :         1         1         0         0         0         0
   Bytes :    1.04 m    1.04 m         0         0         0         0
   Times :   0:00:00   0:00:00                       0:00:00   0:00:00
   Ended : 10 March 2024 19:25:24
```

PCAssistSoftware:
If you can confirm a) whether I need to change it to a capital M as above, and b) that you are happy with this - then I will get this one merged in for now at least

RFBomb:
The case does not matter as we are looking at the comparison using IgnoreCase.

I'm vaguely recalling that the Upper/Lower in the log itself indicates why robocopy did with the file. Upper is include, lower is skip. Maybe vice versa. But we are handling that based on the parameters we know we passed into the command.

PCAssistSoftware:
idiot (me not you) - I should have checked the code and realized we were using CurrentCultureIgnoreCase - sorry been a long day, not thinking fully, apologies.

Yes that is correct, just found my notes on it, uppercase means copied, lowercase means skipped for the following:

Newer / newer
Older / older
Changed / changed
Tweaked / tweaked
Same / same

Are you happy for me to merge this then?

RFBomb:
Yep look good

PCAssistSoftware:
@RFBomb 

Quick question.

Should we have also included "LogParsing_ModifiedInclusion" in the following locations:

ProcessedFileFlags.cs > public enum ProcessedFileFlag

and

RoboSharpConfiguration.cs > public static string GetFileClass

RFBomb:
Yea, good catch.

PCAssistSoftware:
> Yea, good catch.

will get it done shortly then

PR #202, opened 2024-03-18
Memory Usage Optimization

RFBomb:
This PR is meant to resolve some memory usage scenarios that are effectively a memory-leak during long-running operations.

Old Functionality : 
- ResultsBuilder records every line into a List<string>

New Functionality : 
- Added `EnableFileLogging` property to the RoboSharpConfiguration object
- ResultsBuilder functionality changed according to the above property 
      - New default functionality records only the header and summary into the List<string>
      - When `EnableFileLogging` is set true, records all lines.
      - Added a Queue<string> object which is now used for summary parsing, instead of the List<string> object.
- RoboCommand can now have the Configuration updated via a setter
- RoboCommandFactory now has a Configuration property and applies it to generated commands.

### DIFF
--- a/RoboSharp.BackupApp/JobHistoryExpander.xaml
+++ b/RoboSharp.BackupApp/JobHistoryExpander.xaml
@@ -18,12 +18,12 @@
             <RowDefinition Height="*"/>
         </Grid.RowDefinitions>
         <Grid.ColumnDefinitions>
-            <ColumnDefinition Width="300"/>
+            <ColumnDefinition MinWidth="300" MaxWidth="600"/>
             <ColumnDefinition Width="*"/>
         </Grid.ColumnDefinitions>
         <!-- Header Label, ListBox, RemoveBtn -->
         <Label Name="DescriptionLbl" Content="This contains a list of the results from all previous runs during this session." Grid.Row="0" Grid.ColumnSpan="2" HorizontalAlignment="Left" Margin="10,10,0,0" VerticalAlignment="Top" Width="669"/>
-        <ListBox x:Name="ListBox_JobResults" Grid.Row="1" Grid.Column="0" Grid.RowSpan="2" HorizontalAlignment="Left" Height="320" Margin="10,0,0,0" VerticalAlignment="Top" Width="282">
+        <ListBox x:Name="ListBox_JobResults" Grid.Row="1" Grid.Column="0" Grid.RowSpan="2" HorizontalAlignment="Left" Height="320" Margin="10,0,0,0" VerticalAlignment="Top">
             <ListBox.ItemTemplate>
                 <DataTemplate>
                     <Grid Height="20">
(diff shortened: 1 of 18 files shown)


PCAssistSoftware:
Not sure it is working correctly, or I have missed something.

Checked EnableFileLogging is definitely set to False

image

Ran a quick job initially and looked and can still see all information in LogLines and the summary is appearing twice?

image

PCAssistSoftware:
Thanks - now working fine on quick test.  Will run large job now and monitor the memory usage :)

PCAssistSoftware:
Working like a dream so far, large job has been running 30 minutes, but now we would have been well over 1Gb of memory usage before the changes, currently it is just using 124mb!! :)

 I will let the job run to the end to give us a final comparison of the improvements.

RFBomb:
I'm still astonished that its using 124mb (Ive never actually taken these things into consideration or inspected them), but that is a huge improvement!
I've just added in some logic for the Example App as well to showcase the new functionality.

RFBomb:
This screenshot is of the memory profile after running a job with roughly 10k files. The larger instance is the old LogLines (all files included) and the smaller instance is with only the header/summary being recorded into the log lines.  This was tested using the backup app for profiling.  ( The Process Memory was after about 16 minutes of run time on the app and about 20 different tests before I cleared all results and took the screenshot of the two back-to-back runs ). 
image
.

PCAssistSoftware:
A major improvement - my job has now been running for 168 minutes and actually the memory used according to task manager has dropped from the previous 124mb to just 63.5mb!!   fantastic work with these changes, will make a huge difference when doing large server migrations etc

RFBomb:
Another test run, 90k files. 
Snapshot 1 = After first run, EnableFileLogging = false (new default)
Snapshot 2 = After second run, EnableFileLogging = true ( old functionality )

image

RFBomb:
Alright, I think this satisfies this unless you can locate another memory leak issue, or have some other suggestions. 

As a note for future readers: I let the header and summary lines in the RoboCopyResults.LogLines because I felt that these are the most important bits for a results object. If the file lines are still desired, then should either be recorded via the exposed events of RoboCommand, or via a log file produced by RoboCopy itself. The events declaring when a copy is started/completed still occur, but the log line is no longer held in memory by default, which is where the memory leak came from. But the header and summary are only a few lines by comparison (roughly 16-30), and can provide extra within the results object, for minimal memory impact.

Another setting could be added to configuration for disabling the results object log lines altogether, but this can also be accomplished by using the 'LoggingOptions.NoJobHeader' property.

PCAssistSoftware:
Excellent, thank you Robert, much appreciated.   My previous test is STILL running, been 265 minutes now and still only using 66.2mb of memory!, before changes this would easily have been 4+Gb by now.

Will pull your last 6 commits later and run some more tests, but I definitely think this satisfies the memory leak issues.

Will report back once more tomorrow and if all okay then will get it merged.

RFBomb:
Glad to hear the improvements are that big. 

>Will pull your last 6 commits later and run some more tests, but I definitely think this satisfies the memory leak issues.

The other commits shouldn't have too much of an effect on the memory, they were just doing some cleanup of object references and resolving some items I noticed here and there, as well as putting in a checkbox for re-enabling this in the testing app.

PCAssistSoftware:
I have run several more tests both with the BackupApp and also with my apps which use RoboSharp and without a doubt it is a massive improvement in memory usage in all cases I have run.   

This will definitely help when running large jobs for server migrations as previously it was using up ½ of the RAM of the host system just for storing the LogLines, so the reduction in memory usage will undoubtedly also be reflected in the copy performance from the same server.   

Thanks again.  If you have no objection I will merge it now.

RFBomb:
Merge away

PR #205, opened 2024-03-29
Add /FFT parameter support for FAT file times

HitTheDrum:
This commit introduces support for the /FFT parameter, which assumes FAT file times with two-second precision. This enhancement is crucial for ensuring compatibility and precision in file synchronization tasks, especially when working with file systems that follow the FAT time conventions.

This update addresses the need for finer control over file time precision, making RoboSharp more versatile for users dealing with a variety of file systems.

### DIFF
--- a/RoboSharp/CopyOptions.cs
+++ b/RoboSharp/CopyOptions.cs
@@ -72,6 +72,7 @@ namespace RoboSharp
             EnableRestartMode = copyOptions.EnableRestartMode;
             EnableRestartModeWithBackupFallback = copyOptions.EnableRestartModeWithBackupFallback;
             FatFiles = copyOptions.FatFiles;
+            FatFileTimes = copyOptions.FatFileTimes;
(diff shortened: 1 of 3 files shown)


PCAssistSoftware:
I am confused by your pull request as RoboSharp already has support for the /FFT switch?

https://github.com/tjscience/RoboSharp/blob/016b161e47e7155d12b4d7f3e23e150efbfea825/RoboSharp/SelectionOptions.cs#L99

https://github.com/tjscience/RoboSharp/blob/016b161e47e7155d12b4d7f3e23e150efbfea825/RoboSharp/RoboSharp.xml#L1016-L1021

HitTheDrum:
Oh my, sorry for the mix-up with the `/FFT` parameter in my recent pull request. I totally missed that the feature was already baked in because I searched at the wrong place – my bad. I'm pulling my PR now to keep things tidy. Thanks for your patience, and I'll make sure to double-check next time!

Cheers!

PR #209, opened 2024-05-27
Generated regex - Fix #208

### LINKED ISSUE #208: "The input string 'dddd,ddd' was not in a correct format" exception on version 1.4.2 and higher
Hello

I was using robosharp 1.3.5 with the following code and it was working fine.
I tried to update to the latest version and i get the following exception:

```powershell
[15:28:39 ERR] One or more errors occurred. (The input string '74567.813' was not in a correct format.)
   at RoboSharp.RoboCommand.<>c__DisplayClass117_0.<GetRoboCopyTask>b__1(Task continuation)
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
--- End of stack trace from previous location ---
   at RoboSharp.RoboCommand.StartAsync(String domain, String username, String password)
   at Builder.Helpers.Tools.CopyDirectory(String sourcePath, String destPath, Boolean mirror) in D:\dev\Builder\Builder\Helpers\Tools.cs:line 126
```

I tried switching version until it works again and it seems the issue comes from version 1.4.2

Here's the code i'm using to call robosharp:

```C#
RoboCommand cmd = new();
cmd.CopyOptions.Source = sourcePath;
cmd.CopyOptions.Destination = destPath;
cmd.CopyOptions.CopySubdirectories = true;
cmd.CopyOptions.EnableRestartMode = true;
cmd.CopyOptions.MultiThreadedCopiesCount = Environment.ProcessorCount;
cmd.RetryOptions.RetryCount = 5;
cmd.RetryOptions.RetryWaitTime = 5;
cmd.RetryOptions.WaitForSharenames = true;
if (mirror)
{
    cmd.CopyOptions.Mirror = true;
}
else
{
    cmd.CopyOptions.EnableRestartModeWithBackupFallback = true;
}
RoboCopyResults results = await cmd.StartAsync();
```

Environment.ProcessorCount is usually 20
mirror is set to true in this case

Any clue on what could be the issue?

RFBomb:
Updates the library to use the .Net7+  `[GeneratedRegex]` attribute on potential hot-paths. 
Fixes #208 by first attempting to parse in current culture, and only attempting conversion in invariant culture if the current culture failed.


RFBomb:
@PCAssistSoftware this should be complete, I don't think any other regex need to be updated, but feel free to evaluate.

PCAssistSoftware:
I assume from your description and a quick scan of the code that these changes mostly only affect if using .NET7+.  I only currently use the net452 DLL for my project as it uses .NET 4.8, so not sure I can test much unless I create a new test project quickly.  But more than happy to if you want.

RFBomb:
These changes are primarily optimizations for .Net7 and newer, yes. The regex itself will not have changed in most cases. Targets older (such as net452, NetStandard, etc) may see a minor improvement in some cases (such as no longer creating a new regex object every time some calls occur).

This was more to bring the library closer to in line with the guidelines recommended on the Regex pages. 

The vast majority of, if not all of, these are covered by unit tests. 

Tomorrow I intend to add one more commit, but it's basically done here.

PCAssistSoftware:
Tested the PR with my current apps (all using .net 4.8) and no problems to report.  Created a quick testing app using .Net7 and again all seems to be working as expected.

RFBomb:
Just a quick update on this, I was doing some testing and benchmarking to verify that this is the route that we wanted to go and before commit the additional changes. Once I've reviewed the benchmarks a bit more I will submit the final commit to this which has the various changes. 

I intend to then follow that up another request  to submit the actual benchmarking project and the other changes that I have in queue, such as the changes to deprecate framework 452 and replace with 462(which supports.net standard 2.0)

I was also taking a look at  deprecating now out of support.net5 netcoreapp3.1 and letting their support fall back the net standard 2.1 implementation, I'm actually going to hold off on that because my app is currently locked into .net5 unfortunately, But also, the benchmarks are proving that both of those targets see decent improvements in their benchmarks when compared to the standard 2.1 target (for example .net5 performs the regex in approximately half the time compared to 2.1)

RFBomb:
@PCAssistSoftware  Slew of changes submitted.

- Added some unit tests for regex. 
- Created a Benchmarking project that can be run. I was using this to validate the changes done to the regex, and to check out the various optimizations.
- Added a .Zip file to the collection of files used in unit tests to ensure that the issue with the CodePage(437) is working properly
- Made some optimizations for processing output data from the robocopy process
- Reorganized the CSProj file, and added in the newer targets for NetFramework, in anticipation of deprecating Net452.

Running local tests now (they all passed).

One change I implemented here was for the ResultsBuilder. 
- Original Functionality : All items (files and errors) were reported in the log.
- The introduction of Configuration.EnableFileLogging = false was a memory optimization, but had the side-effect of no longer writing errors into the log!
- This introduced a method to ensure that any errors generated during the copy operation properly get reported in the log. 

Without the addition (and with Configuration.,EnableFileLogging = false):
```

```

With the new addition : 
```
2024/05/30 13:44:51 ERROR 32 (0x00000020) Copying File C:\Repos\RoboSharp\RoboSharpUnitTesting\bin\Release\net8.0\TEST_FILES\STANDARD\4_Bytes.txt
The process cannot access the file because it is being used by another process.
```

@PCAssistSoftware - All unit tests are passing locally for me, so if they work for you as well then this should be mergable (for real this time, unless you find something wrong).

PCAssistSoftware:
Excellent work.

All unit tests passing for me locally as well. :)

Great spot on the errors not being written to log and fixing it.

I will test a bit more tomorrow and then merge it in.

RFBomb:
Sounds good to me! 

While I was taking a look at the regex and running unit tests, some numbers I was seeing felt funny, so I spent a few days learning all about BenchmarkDotNet, tested various things (each benchmark takes roughly 30 min to run or so) to check out what the best ways to write things are. For large files, I don't anticipate these changes have too much of an effect, but for operations that span hours with many smaller, quick files, I would expect it to have slightly less overhead overall.

PCAssistSoftware:
I hadn't come across BenchmarkDotNet before but just been looking at the website and it looks really useful

PCAssistSoftware:
All working fine for me in existing apps etc, so will get it merged in shortly :)  Thanks again for all the work you do.

PR #214, opened 2025-07-05
Upgrade CSWin32 from 0.3.106 to 0.3.183

### LINKED ISSUE #213: Add the LogonType in the RoboCommand.StartAsync()
Hi!

The LogonType in the impersonation logic in the API is hardcoded to LogonType.Interactive.

We are using a logon type **Service** in production and **Interactive** in our sandbox, we have our own advapi32 implementation LogonUser and we have the logon type as parameter. 

RoboSharp alos have it's own implementation of LogonUser but we can't specify the logon type.

Is this possible to add a parameter in the **StartAsync** that will be the logon type ?

Something like that, specifying Interactive as defaut to avoid _breaking changes_

```
public virtual async Task<Results.RoboCopyResults> StartAsync(string domain = "", string username = "", string password = "", LogonType logonType = LogonType.Interactive)
{
    await Start(domain, username, password, logonType);
    return GetResults();
}

public enum LogonType
{
    Interactive = 2,
    Network = 3,
    Batch = 4,
    Service = 5,
    NetworkCleartext = 8,
    NewCredentials = 9,
}
```
Regards

RFBomb:
- This upgrade removed some 'AsSpan' functions, so they were re-implemented within the CSWin32Extensions file, ensuring their availability hence-forth.
- Modified unit tests for symbolic link to always use the library, instead of potentially calling the native functions in .net8
- Added "ALLOW_UNPRIVILEGED_CREATE" as a public static value. The assumption here is that the service or application calling this code either will allow or disallow unprivileged creation across the board.

Hopefully resolves #213 compiler issues. Does not attempt to resolve the adding of the login modes.

### DIFF
--- a/RoboSharp.Extensions.UnitTests/SymbolicLinkTests_RequiresAdminPriviliges.playlist
+++ b/RoboSharp.Extensions.UnitTests/SymbolicLinkTests_RequiresAdminPriviliges.playlist
@@ -1,0 +1,27 @@
+<Playlist Version="2.0">
+  <Rule Name="Includes" Match="Any">
+    <Rule Match="All">
+      <Property Name="Solution" />
+      <Rule Match="Any">
+        <Rule Match="All">
+          <Property Name="Project" Value="RoboSharp.Extensions.UnitTests (net5.0)" />
+          <Rule Match="Any">
+            <Property Name="Namespace" Value="RoboSharp.Extensions.Windows.UnitTests" />
+          </Rule>
+        </Rule>
+        <Rule Match="All">
+          <Property Name="Project" Value="RoboSharp.Extensions.UnitTests (net472)" />
+          <Rule Match="Any">
+            <Property Name="Namespace" Value="RoboSharp.Extensions.Windows.UnitTests" />
+          </Rule>
+        </Rule>
+        <Rule Match="All">
+          <Property Name="Project" Value="RoboSharp.Extensions.UnitTests (net8.0)" />
+          <Rule Match="Any">
+            <Property Name="Namespace" Value="RoboSharp.Extensions.Windows.UnitTests" />
+          </Rule>
+        </Rule>
+      </Rule>
+    </Rule>
+  </Rule>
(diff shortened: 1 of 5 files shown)


PCAssistSoftware:
Hi Robert, thank you so much for taking some time to look at this issue, really appreciate it :)   This works fine for me, so if you are happy I will proceed and merge it in.  Thanks again

RFBomb:
NOTE : The changes in this PR are no longer compatible with CSWin32 0.3.106 due to the API changes for creating symlinks that are source-generated. Namely using the `Fixed(....) {   }` pointer blocks.

This was a BREAKING CHANGE that CSWin32 created. The old code in place would have worked for both the version I specified and the 0.3.162 . But the generated code was changed for 0.3.183. Unfortunately, but thats what you get when source-generating. 

As far as I'm concerned, this is ok to merge, pending your review of course.

RFBomb:
Also note : This ADDS A FEATURE that I caught was added via the new generated code, potentially allowing for creation of symlinks without requiring admin privileges, application dependent. I've put that as a static field in the class, since the app will or will not be running in admin mode, so I figured that a static option was OK, especially since its buried in the class.

Finally, I did not up-rev the version of the package. Doesn't matter to me or not, so feel free to do so if you want it to be put out as a newer nuget package.

PCAssistSoftware:
> Also note : This ADDS A FEATURE that I caught was added via the new generated code, potentially allowing for creation of symlinks without requiring admin privileges, application dependent. I've put that as a static field in the class, since the app will or will not be running in admin mode, so I figured that a static option was OK, especially since its buried in the class.
> 

That makes sense and could definitely prove useful.

> Finally, I did not up-rev the version of the package. Doesn't matter to me or not, so feel free to do so if you want it to be put out as a newer nuget package.

As this fixes an issue mainly related to building of the source then I see no reason to up the package version for now, we can always do later if any other changes made 

I have tested here and all seems in order, so will merge it shortly.  Thanks again

PR #217, opened 2026-03-25
Mocks & New IRoboCommand for non-windows platforms

RFBomb:
I required some mocks for my project, so I opted to add them into the Extensions package.
To make best use of the mocks, I'm also creating a new IRoboCommand to fully simulate the actions of RoboCopy via the IFileCopierFactory interfaces that were introduced a while back. This would allow full control of the copy and logging process by consumers that want it, and should also allow for cross-platform usage via custom IFileCopiers (like the StreamedFileCopier)

# Overview

## Base library changes:

- Exposed some private methods from Authentication.cs
- Added method `RetryOptions.GetRetryWaitTime()` to provide a TimeSpan based on the specified wait time
- Improved documentation comments for SelectionOptions and CopyOptions
- Added ProcessedDirectoryFlag.MisMatch
- Added a storage int for the enum value for `ProcessedFileInfo`

## Add RoboSharp.Extensions.RoboCommandPortable

- This is a new IRoboCommand that is fully managed instead of starting a new process outside of the application's control.
- This utilizes an IFileCopierFactory to allow consumers to create their own implementations, or use the baked in `Windows.CopyFileEx` or `StreamedCopier` factories.
- The aim for this class is parity (as close to it) with RoboCopy, while also bringing support to cross platform targets.
- Note : RoboCommand is still the suggested way to go due to speed and the backing of Microsoft, but this allows for more power over unit tests if (which was my requirement) and also it should work fine in a linux environment. I can't vouch for speed compared to other tools in that space though.
 

## RoboSharp.Extensions.Mocks

- This namespace has been introduced for various mocking for use with Dependency Injection unit testing where you may not want to actually perform copy operations during the test, but still need an IRoboCommand.

## Unit Testing

- Many unit tests have been refactored to make the tests more targeted and more consistent. No more sharing of destination directories between tests. What was being tested has not changed, but the structure of the tests have. 
- CommandTests<T> has been introduced to learn how robocopy works and validate the tests against actual robocopy.
- RoboMover and RoboCommandPortable tests are now derived from this test suite to ensure they perform identical to robocopy.

## BackupApp

- Removed Net48 target in favor of .Net10-windows and to allow testing of the additional command types provided by RoboSharp.Extensions
- Now uses an IRoboCommandFactory to create the command of the specified type. Defaults to standard RoboCopy.
- Added support for LoggingOptions logging paths

## Build-And-Test Github Action

- A github action has been introduced to build the solution, run all robosharp tests, then run all extensions tests, and finally grab various artifacts from the build.
- The action will run on pushes to dev and pull requests to dev.


PCAssistSoftware:
nice, I will keep following the progress of these changes. Thanks Robert

RFBomb:
Just an update to this, I've added a github action to my repo. It does basically the same thing as AppVeyor integration, but I find it provides more useful and more controlled behavior. 

I have it set up to build the solution, run the primary tests (against Robocommand), then run the extension tests. If all completes, it will package up the nuget packages and the compiled exe's (for backup app and console app) and upload as an artifact. 

This can also be set up to deploy to nuget automatically on a successful push to main (dev) if you wanted to implement that. 

I'm also reworking some unit tests to be more efficient, more isolated, and better for testing custom implementations (RoboMover and the new RoboCommandPortable which is a managed version of RoboCopy instead of calling a process). 

Note i do not plan to implement some advanced functions (scheduling, useFAT times, etc), but core functionality should be 1:1.

PCAssistSoftware:
Okay thanks for update.

The idea of using github actions instead of appveyor is interesting and something I may consider at some point, but for now quite happy with appveyor and they have been very helpful recently in adding the VS 2026 and .NET 10 images for me to use for the most recent release.

RFBomb:
I just wanted to inform you of the addition. Something I’ve noticed is that AppVeyor is passing while the GitHub action is (correctly) failing the unit tests.

PCAssistSoftware:
I appreciate it, its certainly good to know of an alternative, as appveyor can sometimes be really problematic, and slow in supporting new .NET versions etc.  So thank you

RFBomb:
Ok, I believe I'm done with this. 
The only _semi-major_ thing I haven't implemented for the new IRoboCommand is the average speed. If people have a desire for that, that can be done later. 

image

The new command can be tested out via the BackupApp. Based on some cursory testing, they are slower, since I don't have the power of Microsoft behind me for optimization. But my primary goal wasn't full speed for this pass, it was just getting an implementation and reworking the unit tests. I wouldn't use it for cross-server anything (robocopy will be faster), but it should allow experimentation for linux and docker images for applications that want to try it.

PCAssistSoftware:
Okay, I will give it a try out when I get a moment and then get back to you. Please bear with me though as currently stacked with work before going away for a few days, so it maybe when I get back

RFBomb:
No worries, I'm in absolutely no rush for this. I was more so finishing the original idea behind the extensions than anything, as well as implementing some nice-to-haves for unit testing that I found myself needing.

PCAssistSoftware:
Hi

Finally found some time to take a look at this.

First question, why the change below as this breaks existing functionality in my apps?

RoboSharp/DefaultConfigurations/RoboSharpConfig_EN.cs      line 36

            LogParsing_ExistingDir = "Existing Dir";

changed to

            LogParsing_ExistingDir = ""; 

And also related area to this e.g.

RoboSharp/RoboSharpConfiguration.cs    line 329

           get { return existingDirToken ?? GetDefaultConfiguration().existingDirToken ?? "Existing Dir"; }

changed to

            get { return existingDirToken ?? GetDefaultConfiguration().existingDirToken ?? ""; }

PCAssistSoftware:
any other breaking changes I need to be aware of?

RFBomb:
Good catch, I'll re-review. That should be the only one, but I agree with you and I'll revert it since its been there a long time.

Robocopy doesn't emit a token there for existing dirs, so it was breaking my unit tests, if i recall correctly. But thats easy enough to handle with a custom configuration, so that is what I'll implement for the unit tests.

RoboSharp/RoboCommand.cs
```diff
@@ -25,6 +25,11 @@ public partial class RoboCommand : IDisposable, IRoboCommand, ICloneable
         /// </summary>
         public static RoboCommandFactory Factory { get; } = new RoboCommandFactory();
 
+        /// <summary>
+        /// char array used for splitting log lines for files and direcetory info
```
PCAssistSoftware: typo: direcetory should be directory

RoboSharp/RoboSharpConfiguration.cs
PCAssistSoftware: Believe this may need to go back to

get { return existingDirToken ?? GetDefaultConfiguration().existingDirToken ?? "Existing Dir"; }

as it was before??

PCAssistSoftware (dismissed):
couple of comments left in the code, will test fully later

N.B. First time using submit review, so hopefully this works!

RoboSharp/RoboSharpConfiguration.cs
RFBomb: Good catch, I missed that one since I focused on the _EN implementation.

PCAssistSoftware (commented):
All looking good, the initial problem I encountered is now resolved.   Will run some more thorough tests later today or tomorrow and then get back to you, but unless I find any issues with existing apps this should be good to merge

RFBomb:
OK cool, I'm just did one last commit to normalize the version numbers, since I noticed I bumped extensions greater than robosharp. Both will be v1.6.3 (nuget is currently 1.6.2). 

(Extensions had been bumped more due to my messing around with an local nuget repository, but once its published I'll defer my stuff back to standard nuget)

PCAssistSoftware:
Okay, have now tested this release with all my apps and is working fine, also ran all the tests and had a look and test of new functionality via backup app and all looks good, so if nothing more you want to amend then I am more than happy to merge.  Thanks again for all you do.

RFBomb:
I'm happy that its all working. As part of this commit I had buffed up some of the unit tests as well to be better testing of some things and allow easier testing of others. Part of that was isolating each test to a randomly generated folder instead of always using same source and destination, which previously (rare occasion) had some issues polluting between tests. 

Suffice to say, merge away!

